PR #73, opened 2021-09-26
Click posts while in pan mode

### LINKED ISSUE #70: Edit/click posts while in pan mode
Be able to click on and edit/delete posts while we are in panning mode.

Note: this doesn't mean we can move them, only click on them

Ammar-T:
- Able to click on posts in any mode

Closes #70


JoelWiebe (approved):
Great work! I noticed a bug with drag-note mode so I'll create that as a separate issue, but pan mode looks good.

PR #84, opened 2021-10-10
Only allow images for canvas bg

### LINKED ISSUE #80: Restrict upload background file types
From the board configurations (and new board creation), restrict the background image file types that can be uploaded to image file types (e.g., .jpg, .JPEG, .png, etc.) and do not allow selection of other file types (e.g., PDF, .doc, etc.).

There should be an option to restrict the file types that the user can select.

Ammar-T:
- Allowing .jpg, .jpeg, .png, .svg

Closes #80


JoelWiebe (approved):
Tested uploading of
- svg
- png
- jpg

Although they allowed me to upload, I was getting errors and the png and jpg did not present on the screen. I'll add this as a separate issue.

PR #85, opened 2021-10-11
Added submit on enter + fixed password manager issues

### LINKED ISSUE #78: Rename create account fields to support password managers
Password managers (such as the integrated Chrome password manager) erroneously saves the "username" field and the "password" field, however, the "email" and "password" fields are used for login.

Could we change the username field to simply "name" or "Preferred name" or something that does not imply that it should be used for logging in? Also, could you test to make sure that Google password manager or others grab the email instead of the name when autosaving the password (I'm not sure how they go about locating the appropriate field)?

Ammar-T:
- For register and login, hitting enter on keyboard will submit now
- Password manager issue should be fixed, it works for the password manager I use, @JoelWiebe can you verify it works with the Chrome password manager as well?

Closes #78

### DIFF
--- a/src/app/components/login/login.component.html
+++ b/src/app/components/login/login.component.html
@@ -1,7 +1,7 @@
 <div class="wrapper">
     <div class="container">
-        <img src="../../assets/register.png" />
-        <div class="formContainer">
+        <img src="../../assets/Login.png" />
+        <form class="formContainer">
             <div>
                 <h1 class="header">Welcome Back!</h1>
                 <p class="subheader"></p>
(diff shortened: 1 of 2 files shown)


Ammar-T:
> For some reason the Google password manager is still defaulting to selecting the top field as the username. Although Google allows people to select the email as the second option, not everyone would know that: image
> 
> I also couldn't find anywhere a description of how Google identifies the fields... maybe there are some Google developer docs for their registration field standards/recommendations somewhere?

So you can add some attributes to html tags which would identify certain inputs to be "username", "email", "password", and so on. I was initially tagging email as "email" and username as "username". But since Google's manager uses "username" to save the email/username, I tagged the email with "username" instead. So, it works now haha.

PR #93, opened 2021-10-27
Tagging V1

Ammar-T:
Implemented tagging within a post's modal.
- Students can add or remove tags from inside a post's modal
- Teachers can add or remove tags from the settings modal

For V2, we should probably show a limited number of tags on the actual post itself as well.

### DIFF
--- a/src/app/components/add-board-modal/add-board-modal.component.html
+++ b/src/app/components/add-board-modal/add-board-modal.component.html
@@ -24,6 +24,22 @@
         <mat-checkbox [checked]="allowStudentMoveAny">Students can move any post?</mat-checkbox>
       </p>
     </mat-tab>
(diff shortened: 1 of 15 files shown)


Ammar-T:
@JoelWiebe Added your 2 suggestions! 

Some of the big features like likes, comments, and tagging are hard to make backwards compatible so unfortunately tagging will only work on posts created after this PR is merged.

JoelWiebe (approved):
Tested Add Tag:
- Enabled when tag is being written
- Greyed out when hit enter or clicked Add Tag

Tested creating tag:
- Tag saves as soon as enter or Add Tag are pressed (no need to click Update Configuration)

Tested tagging post:
- Tags can be added from the post modal pop-up

Issue #67, opened 2021-09-16
Add option to lock student editing


JoelWiebe:
From teacher's board configurations, add a separate options to lock students from (1) creating/editing, (2) commenting, (3) voting, (4) tagging posts on the CK Board. Users can still open posts to view them while all locks are enabled. Additionally, these configurations apply only to the CK Board (i.e., even when locks are enabled for the board, students would still be able to edit from the CK "Bucket").

**Acceptance Criteria:**

- In the `configuration-modal`, add a toggle setting that would lock creating, commenting, liking, tagging posts
- In the `add-board-modal`, add a toggle setting that would lock creating, commenting, liking, tagging posts

JoelWiebe:
Thanks @Ammar-T for updating the issue description--I just added another updated that all features get separate locks (i.e., creating/editing, voting, commenting, tagging can each be individually locked to provide the most flexibility) ... I'll mark the other issue related to this as a duplicate

PR #105, opened 2021-11-12
added option to lock student editing

l3n0ire:
added option to lock student editing
fixed bug with allowStudentMoveAny not updating correctly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,17285 @@
 {
   "name": "ck-board",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "ck-board",
+      "version": "0.0.0",
+      "dependencies": {
+        "@angular/animations": "~12.0.1",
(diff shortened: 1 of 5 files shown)


Ammar-T:
Can you remove the `package-lock.json`, since the 17,000+ changes doesn't seem right haha.

Ammar-T:
Also, I should've been clearer but the branch name should be something like `issue-67` and put `Closes #67` in the PR description so it can link this PR with the issue.

JoelWiebe (changes requested):
Tested
- When locked, teacher can still create/edit
- When locked, student cannot create/edit
- When unlocked, student can create/edit again

Noticed that student must refresh the browser before changes take effect, this should be fine for the time being. 

For consistency of permissions settings, should we be saying: "Students can move posts"; "Students can create, edit, or delete posts"

Right now with one question and one statement, it's a bit confusing. What do you think about the exact phrases above?

Ammar-T:
@l3n0ire I meant remove the package-lock changes from the PR not the repository. `package-lock` should generally be committed and not ignored in Git repos.

PR #106, opened 2021-11-21
ISSUE-97: Adds private/public boards with join codes

### LINKED ISSUE #97: Create Permissions to Scope Access to Boards
Boards should be scoped to only be visible in dashboards to teachers or students who are subscribed to that board. We may also consider access permissions such as private, public, and hidden/unlisted for each board.

Ammar-T:
#### New Features:
- When creating or editing board settings, you can choose to make board `public` or `private`
- Dashboard is split into 2 sections now: `Your Boards` and `Public Boards`
- Dashboard navigation bar has new button which allows user to join new boards using an invite code

Closes #97

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -2,6 +2,7 @@ import { NgModule } from '@angular/core';
 import { RouterModule, Routes } from '@angular/router';
 import { CanvasComponent } from './components/canvas/canvas.component';
 import { DashboardComponent } from './components/dashboard/dashboard.component';
+import { ErrorComponent } from './components/error/error.component';
 import { LoginComponent } from './components/login/login.component';
 import { RegisterComponent } from './components/register/register.component';
 import { AuthGuard } from './utils/auth.guard';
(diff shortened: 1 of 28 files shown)


JoelWiebe (changes requested):
First of all, fantastic work!

Tested:
- creating a new private board shows up in "Your Boards"
- creating a new public board shows up in both "Your Boards" and "Public Boards"
- tested student joining to add private board to "Your Boards"
- changing a board from private to public will cause other users to see the board in "Public Boards"
- A non-owner teacher can change settings including making the board private (causing themselves to lose complete access to the board -- perhaps this is fine)

The public/private distinction works well. I notice that currently boards private boards aren't restricted, in the sense that if I have the URL I don't need the join code. Maybe we can address this as a lower priority issue. Maybe they get a blank page with a prompt to request join code for access.

Suggested changes:
- Remove case sensitivity (i.e., perhaps make join code all upper/lowercase and remove case sensitivity when inputting the code -- perhaps this is a non-issue as it was one of our older boards that had mixed cases and all new boards were all uppercase)
- In join code, would it be possible to prohibit confusing characters (i.e., "l" and "o" that look similar to numbers) ... maybe this and the above item can be part of a new issue
- Students cannot create boards (maybe we can do this as an additional issue -- whatever you think is best!)
- Change "Users" tab and "Current Users" label to "Members" and "Current Members" 
- Change "Your Boards" to My Boards in the dashboard

Let me know your thoughts regarding the above and, again, great work!

PR #111, opened 2021-12-01
Feat/103 password recovery

### LINKED ISSUE #103: Add email-based password recovery
- Add "forgot password" option on login page
- Show error to user if email does not exist
- Notify usr if email was sent
- Include a link in an email to allow the user to reset and confirm their new password

StevenWuzz:
## Details

Added email-based password reset/recovery:

- A new page for requesting password recovery via email
- Send an email containing a link to reset the user's password
- A new page to enter the new password and confirm it

Closes #103

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -3,6 +3,7 @@ import { RouterModule, Routes } from '@angular/router';
 import { CanvasComponent } from './components/canvas/canvas.component';
 import { DashboardComponent } from './components/dashboard/dashboard.component';
 import { LoginComponent } from './components/login/login.component';
+import { PasswordResetComponent } from './components/password-reset/password-reset.component';
 import { RegisterComponent } from './components/register/register.component';
 import { AuthGuard } from './utils/auth.guard';
 import { BoardGuard } from './utils/board.guard';
@@ -15,6 +16,7 @@ const routes: Routes = [
     canActivate: [AuthGuard]  
   },
   { path: 'login', component: LoginComponent },
+  { path: 'password/reset', component: PasswordResetComponent},
   { path: 'register', component: RegisterComponent },
   { path: '**', redirectTo: 'login' }
 ];
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -16,7 +16,7 @@ const routes: Routes = [
     canActivate: [AuthGuard]  
   },
   { path: 'login', component: LoginComponent },
-  { path: 'password/reset', component: PasswordResetComponent},
+  { path: 'reset-password', component: PasswordResetComponent},
   { path: 'register', component: RegisterComponent },
   { path: '**', redirectTo: 'login' }
 ];
--- a/src/app/components/login/login.component.html
+++ b/src/app/components/login/login.component.html
@@ -18,6 +18,7 @@
                     </mat-form-field>
                     <p *ngIf="invalidCredentials" class="invalid-pass">Invalid email or password!</p>
                 </div>
+                <a routerLink="/password/reset" id="request-pass-reset">Forgot Password?</a>
                 <div class="next" >
                     <button type="submit" mat-button>Login</button>
                 </div>
--- a/src/app/components/login/login.component.html
+++ b/src/app/components/login/login.component.html
@@ -18,7 +18,7 @@
                     </mat-form-field>
                     <p *ngIf="invalidCredentials" class="invalid-pass">Invalid email or password!</p>
                 </div>
-                <a routerLink="/password/reset" id="request-pass-reset">Forgot Password?</a>
+                <a routerLink="/reset-password" id="request-pass-reset">Forgot Password?</a>
                 <div class="next" >
                     <button type="submit" mat-button>Login</button>
                 </div>
--- a/src/app/components/password-reset/password-reset.component.ts
+++ b/src/app/components/password-reset/password-reset.component.ts
@@ -95,6 +95,7 @@ export class PasswordResetComponent implements OnInit, OnDestroy{
         this.auth.getAuth().confirmPasswordReset(this.oobCode, this.newPassword)
         .then(res => {
             this.passwordMsg = "Success";
+            alert("Your password has been successfully updated");
             this.router.navigate(['/login']);
         })
         .catch(() => this.passwordMsg = "Error");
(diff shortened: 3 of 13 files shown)


src/app/app-routing.module.ts
```diff
@@ -15,6 +16,7 @@ const routes: Routes = [
     canActivate: [AuthGuard]  
   },
   { path: 'login', component: LoginComponent },
+  { path: 'password/reset', component: PasswordResetComponent},
```
Ammar-T, line 19:
```suggestion
  { path: 'reset-password', component: PasswordResetComponent},
```
Would prefer to keep it as a single path without extra slashes.

src/app/components/login/login.component.html
```diff
@@ -18,6 +18,7 @@ <h1 class="header">Welcome Back!</h1>
                     </mat-form-field>
                     <p *ngIf="invalidCredentials" class="invalid-pass">Invalid email or password!</p>
                 </div>
+                <a routerLink="/password/reset" id="request-pass-reset">Forgot Password?</a>
```
Ammar-T, line 21:
```suggestion
                <a routerLink="/reset-password" id="request-pass-reset">Forgot Password?</a>
```

src/app/components/password-reset/password-reset.component.ts
```diff
@@ -0,0 +1,110 @@
...
+            this.oobCode = params['oobCode'];
+
+            if(this.mode == "resetPassword") {
+                this.auth.getAuth().verifyPasswordResetCode(this.oobCode)
+                .then(() => {
+                    this.oobCodeChecked = true;
+                })
+                .catch(e => {
+                    alert(e);
+                    this.router.navigate(['/password/reset']);
```
Ammar-T, line 51:
```suggestion
                    this.router.navigate(['/reset-password']);
```

src/app/services/auth.service.ts
```diff
@@ -79,4 +79,13 @@ export class AuthService {
...
     })
   }
+
+  getAuth() { 
+    return this.auth; 
+  } 
+
+  resetPassword(email: string) {
+    return this.auth.sendPasswordResetEmail(email, 
+      { url: 'http://localhost:4200/password/reset' })
```
Ammar-T, line 89:
```suggestion
      { url: 'http://localhost:4200/reset-password' })
```

Ammar-T:
@JoelWiebe Try updating this branch (`git pull origin feat/103_password-recovery`) and then running `ng serve` again. It should work now.

JoelWiebe (changes requested):
This feature looks great! Just a few things that I am noticing. 

There seems to be an issue when opening the link from my email, I am sent to a server error 500 page
image

Also, two small items, could we centre-align this text: "A link to reset your password has been to your email" 
image

And in the email is it possible to replace "project-473282317820" with "Common Knowledge Board"?

Great work, hopefully just some quick adjustments!

StevenWuzz:
I believe the error is most likely caused by URL change, which affects the URL set up in the firebase. It will be a quick adjustment, but currently I am not able to log into the firebase account

StevenWuzz:
@JoelWiebe I believe that I have resolved all the errors and made the adjustments you requested. Please let me know what you think.

JoelWiebe:
@StevenWuzz I notice that I can click submit twice which generates an error. What do you think about the following changes to make this even more user friendly:

- Move the password has been updated message from the pop-up box to in-page text
image

- Somehow prohibit users from clicking the submit button twice (i.e., hiding/greying-out the submit & cancel buttons after clicking submit)
- Somehow adding a button/hyperlink that takes users to the login page so users don't end up stuck on this page -- in whatever way you think makes sense

image

StevenWuzz:
@JoelWiebe 

"Somehow prohibit users from clicking the submit button twice (i.e., hiding/greying-out the submit & cancel buttons after clicking submit)"

Are you referring to both the email entry section and the password reset section as well?

JoelWiebe:
@StevenWuzz The double submit was just an issue for me on the password reset page. 

Automatic redirect is a great idea too. Whichever you think is better in this situation: automatic redirect when the user clicks "OK" in the pop-up OR provide them a "Login" or "Home" button on the page after they update their password. The choice is yours!

StevenWuzz:
@JoelWiebe have you pulled the latest changes? As far as I know, that error on the password reset page is not supposed to happen. After successfully updating the password, the user is supposed to be prompted to click the "OK" button on the pop-up and then redirected automatically to the login page. Hence, it should not be possible to click on the submit button twice, actually.

Ammar-T:
Can we get rid of the popup alert and use an internal popup modal instead? For example, when we click on a post on the canvas we have our own modal from Material-Angular. Reason I say this is because the current popup seems a little old-fashioned 😄  @StevenWuzz

src/app/services/auth.service.ts
```diff
@@ -80,4 +80,13 @@ export class AuthService {
...
     })
   }
+
+  getAuth() { 
+    return this.auth; 
+  } 
+
+  resetPassword(email: string) {
+    return this.auth.sendPasswordResetEmail(email, 
+      { url: 'http://localhost:4200/reset-password' })
```
Ammar-T, line 90: This wouldn't work in production since there's no localhost:4200, can you find another way to route properly? Thanks!

StevenWuzz: @Ammar-T what do you mean by no localhost:4200 though?

StevenWuzz:
> Can we get rid of the popup alert and use an internal popup modal instead? For example, when we click on a post on the canvas we have our own modal from Material-Angular. Reason I say this is because the current popup seems a little old-fashioned 😄 @StevenWuzz

Could you elaborate where I can find that internal popup modal?

Ammar-T:
> > Can we get rid of the popup alert and use an internal popup modal instead? For example, when we click on a post on the canvas we have our own modal from Material-Angular. Reason I say this is because the current popup seems a little old-fashioned 😄 @StevenWuzz
> 
> @Ammar-T Could you elaborate where I can find that internal popup modal?

https://material.angular.io/components/dialog/overview

JoelWiebe (approved):

JoelWiebe:
@l3n0ire Do you mind reviewing this PR (checkout changes and add your review decision: accept/request changes)?

l3n0ire (approved):

PR #114, opened 2021-12-08
Feat/58_zoom-depth-tool

### LINKED ISSUE #58: Create Zoom Depth tool
In bottom right corner of canvas, display the zoom depth and perhaps provide clickable options to zoom in/out or return to 100% zoom

StevenWuzz:
New feature: zoom depth tool via buttons

## Details
-  The zoom point is set to be at the center of the canvas since there's no way to dynamically specify it 
-  The proportion goes up or down by 0.05 per click 
-  The recycle button reset the zoom so that it is 100%.

Closes #58

### DIFF
--- a/src/app/components/canvas/canvas.component.html
+++ b/src/app/components/canvas/canvas.component.html
@@ -39,6 +39,7 @@
         <div class = "toolField"> 
           <button (click) = "handleZoomIn()"><i class="fas fa-search-plus"></i></button>
           <button (click) = "handleZoomOut()"><i class="fas fa-search-minus"></i></button>
+          <button (click) = "handleZoomReset()"><i class="fas fa-recycle"></i></button>
         </div>
       </div>
   </div>
--- a/src/app/components/canvas/canvas.component.html
+++ b/src/app/components/canvas/canvas.component.html
@@ -37,9 +37,9 @@
         </div>
       <div class="zoom">
         <div class = "toolField"> 
-          <button (click) = "handleZoom('zoomIn')"><i class="fas fa-search-plus"></i></button>
-          <button (click) = "handleZoom('zoomOut')"><i class="fas fa-search-minus"></i></button>
-          <button (click) = "handleZoom('reset')"><i class="fas fa-recycle"></i></button>
+          <button (click) = "handleZoom('zoomIn')" class = "zoomButton"><i class="fas fa-search-plus"></i></button>
+          <button (click) = "handleZoom('zoomOut')" class = "zoomButton"><i class="fas fa-search-minus"></i></button>
+          <button (click) = "handleZoom('reset')" class = "zoomButton"><i class="fas fa-recycle"></i></button>
         </div>
       </div>
   </div>
--- a/src/app/components/canvas/canvas.component.html
+++ b/src/app/components/canvas/canvas.component.html
@@ -37,9 +37,9 @@
         </div>
       <div class="zoom">
         <div class = "toolField"> 
-          <button (click) = "handleZoomIn()"><i class="fas fa-search-plus"></i></button>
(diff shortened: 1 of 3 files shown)


StevenWuzz:
Need to resolve the small issue with zooming coordinates.

JoelWiebe:
@StevenWuzz is this ready to review?

StevenWuzz:
@JoelWiebe I still haven't really figured out a way to resolve the issue, unfortunately. Hopefully I can get it done by the time we have our standup this week.

StevenWuzz:
@JoelWiebe I think you can try to review it now. It's far from perfect, but I have consulted with Ammar and at least for now, this is probably the best we can do.

JoelWiebe:
@StevenWuzz Great work, thanks for your time figuring this out. 

I like this solution of zoom in and out to the centre of the canvas. Let's keep that. 

There are a couple things I noticed, please let me know if my proposed solutions may be possible.

Functionality:
- Is it possible to display the numerical value of the zoom depth (e.g., 75%), updating when changes are made to either the zoom tool or mouse zooming? 
- Can clicking on the percentage number zoom the user to 100% (if so, some kind of hover colour change or tooltip might be helpful in indicating that this number can be clicking)?
- It appears that the zoom in/out controls save their own value but do not update to the zoom depth controlled by the user's mouse; if I open a board, zoom in very close using my mouse, then click the zoom in icon I actually zoom out not in. Is there a way for the zoom depth value used by the zoom tool to update with the zoom depth value adjusted by a user's mouse?

The image below illustrates 

- replacement of icons with text: (1) en dash (–) -->note, this is longer than a hyphen; (2) percentage (e.g., 100%), and; (3) plus (+). 
- either a tooltip or changing text colour can be used to illustrate that each item is clickable (i.e., change colour or add tooltip when hovering on 100%)
- I also set the transparency of the white background to 75% to make this more subtle and made small changes to pad this away from the edge and created slightly sharper corner edges

image

Alternatively, do you think we should just remove the + and - icons and only have the percentage displayed if this makes implementation easier? I have displayed this below:
image

Let me know what you think!

Ammar-T:
@JoelWiebe I like those suggestions! 

@StevenWuzz There's material components you can use to get some of those requirements done. 
(mat-tooltip, mat-button)

JoelWiebe:
@Ammar-T Thanks for the added info regarding material components!

@StevenWuzz Sorry, you had already fixed that mouse zoom bug I mentioned -- so you can ignore that functionality bug that I mentioned; when I checkout out the code I hadn't pulled the latest commit!

JoelWiebe:
@StevenWuzz Great work this is looking really good!

Just noticed something really small. When resetting to 100% then hitting the zoom-out button twice, it zooms to 95% then to 89%. Just curious why it switches from to 89% instead of 90%. Is there a way to get it to move up or down consistently by an increment or decrement of 5%?

StevenWuzz:
That's apparently a floating-point error in Javascript. In JS, 0.95 - 0.05 = 0.899999...9 and 1.1 + 0.05 = 1.1500000..1. I'm still thinking about how to rectify this

Ammar-T:
@StevenWuzz Can we just round to nearest tenth every time, considering delta is 0.05?

StevenWuzz:
@Ammar-T Ah, yeah, there's a function for that. Yep, solved

JoelWiebe (approved):
Looks great! Thanks @StevenWuzz

JoelWiebe:
@l3n0ire @Ammar-T Hey, can you review when you get a chance?

Ammar-T (approved):
👍

l3n0ire (approved):

PR #130, opened 2022-01-09
ISSUE-66: Creating CK-Buckets

### LINKED ISSUE #66: Create CK Bucket page that lists posts from a CK Board
Create a page that can display all contents of the CK Board but displays posts in a list rather than a board. Later, specifications will elaborate ways to display different subsets of a CK Board to different students using this CK Bucket.

Ammar-T:
## Details

- 

Closes #66


JoelWiebe (changes requested):
@Ammar-T this is so awesome! It really extends new possibilities :)

Tested:
- Opening list view to display posts
- Opening buckets view to open posts and vote
- Adding post to a bucket
- Creating a new bucket

Suggestions:
- When opening the list view, each post appears to have the board ID displayed just below the post title (e.g., WnuRWhzEuCX....); can this board ID be replaced with the author name (also would this change when the hide author ID option is selected?) -- [low priority]
- If I open the list view then buckets view, the buckets are hidden behind the list view; can we close the list view when the bucket view is opened (and vice versa)? -- [medium priority]
- Upvoting a post on the board does not immediately upvote the post in the list view or bucket (need to close then re-open the list view or bucket to update -- [low priority]
- Creating a new bucket does not immediately update in the bucket view's bucket list -- [low priority]

In the Workflows tool, I notice that switching between the Workflow and Bucket tabs changes the button labels at the bottom -- very cool. Sometimes it's the small things we appreciate :)

Very, very cool. I just mentioned a few things I noticed as I was navigating through. All the functionality looks great!

Ammar-T:
@JoelWiebe Fixed the first 2 suggestions. 

The latter 2 would require the bucket and list view to be real-time like the board itself, which is a little bit more work so I'll separate it into its own issue. Also, just wondering if making the bucket and list view real-time is 100% necessary since the board already has that capability?

JoelWiebe:
@Ammar-T I'm really liking these features... including the animation of the list and bucket views sliding in and out, and being able to close the views by re-clicking the button. Fantastic. 

Good point above, not making the buckets live. I notice that when the posts in the list view are opened up, the latest contents are displayed... so that should serve most needs. I did notice something interesting though. If I close the modal post pop-up from the list view by clicking the "close" button, the list view is updated, but if I click on the canvas to close the list view modal post pop-up then the list view does not update.

Could we make it so that clicking outside the list view modal post pop-up also updates the list view post, then forget about making the lists and buckets update in real-time?

Ammar-T:
@JoelWiebe Wow, I'm surprised you caught that 😆 . All fixed now!

JoelWiebe:
Thanks @Ammar-T!!! Hahaha, just doing my job ;)

JoelWiebe:
@Ammar-T Could we remove the Workflows tool from the student view?

Ammar-T:
@JoelWiebe Fixed your last suggestion, is this good to merge now?

JoelWiebe:
@Ammar-T I just found two new puzzles to solve :)

Double vote hack!
(1) Open the list view
(2) Like a post on the canvas
(3) Like the same post in the list view
(4) The post now just received 2 likes from the same individual

I suppose there are two possible solutions, either: (1) update the list view when a user likes a post or (2) when a user likes from the list view, verify whether the DB value matches the list view value. P.S. The converse might also be true that too many points are removed if someone removes their like from the canvas then removes their like from the list view. Please feel free to do whatever you think is best!

The other item is that anonymity permissions for the canvas currently don't apply to the list view (cannot anonymize the list view). I'm happy to push this as-is and deal with this bug in another issue so we can get these features on the develop branch and on the server for testing!

PR #135, opened 2022-01-19
Feat/92 additional pan controls

### LINKED ISSUE #92: Additional Pan Controls
Can we add additional pan controls to make it easier to navigate a canvas (i.e., using arrow keys to pan up/down and right/left)? 

Also, is it difficulty to have separate touch events for trackpads vs mouse (i.e., using two finger swipe on a trackpad to move up/down and right/left)?

StevenWuzz:
## Details

- Additional pan controls by using the arrow keys (up/down/left/right)

Closes #92

### DIFF
--- a/src/app/components/canvas/canvas.component.html
+++ b/src/app/components/canvas/canvas.component.html
@@ -16,7 +16,7 @@
         <button mat-menu-item (click)="this.authService.signOut()">Log Out</button>
       </mat-menu>
   </mat-toolbar>
-  <div class="canvas-area">
+  <div class="canvas-area" id="canvas-area">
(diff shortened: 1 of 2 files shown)


JoelWiebe (commented):
@StevenWuzz I notice that pushing the up and down arrows moves the navigation bar up and down a bit. Would it make sense to add 

"event.preventDefault();"

inside each of the four if-statements to prevent default behaviours for up/down/left/right keys?

StevenWuzz:
@JoelWiebe Good catch! Yes, I've just fixed that.

l3n0ire (approved):

PR #136, opened 2022-01-26
ISSUE-89: Image Upload Bugs

### LINKED ISSUE #89: Error uploading images to canvas bg
When uploading images I get an error in the browser console
image

To reproduce:
- I uploaded an svg (which usually works)
- then a png and a jpg (which both throw the error)

The image sometimes flickers on the screen before disappearing

Ammar-T:
## Details

- Added a new `Remove Background Image` button which is self-explanatory
- Restricted image uploads to `.png` and `.svg` since `.jpeg/.jpg` are very large when encoded to strings

Note: In the future, we should look to move to alternate image storing service which will allow us to store larger images. Or, find a way to compress `.jpg` images.

Closes #89

### DIFF
--- a/src/app/utils/FabricUtils.ts
+++ b/src/app/utils/FabricUtils.ts
@@ -173,8 +173,8 @@ export class FabricUtils {
           left: vptCoords?.tl.x,
           width: width,
           height: height,
-          scaleX: width / (img.width ?? 0),
-          scaleY: height / (img.height ?? 0)
+          scaleX: width / (img.width ?? 1),
+          scaleY: height / (img.height ?? 1)
         }
     }
 }
--- a/src/app/utils/FabricUtils.ts
+++ b/src/app/utils/FabricUtils.ts
@@ -157,4 +157,24 @@ export class FabricUtils {
         obj.addWithUpdate();
         return obj
     }
+
+    createImageSettings(canvas, img) {
+        let vptCoords = canvas.vptCoords
+        let width = canvas.getWidth()
+        let height = canvas.getHeight()
+
+        if (vptCoords) {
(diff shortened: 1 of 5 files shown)


src/app/utils/FabricUtils.ts
```diff
@@ -157,4 +157,24 @@ export class FabricUtils {
...
+          height = Math.abs(vptCoords.br.y - vptCoords.tr.y)
+        }
+
+        return {
+          top: vptCoords?.tl.y,
+          left: vptCoords?.tl.x,
+          width: width,
+          height: height,
+          scaleX: width / (img.width ?? 0),
+          scaleY: height / (img.height ?? 0)
```
StevenWuzz, line 177: Is the division by zero (when img.width or img.height is null or undefined) intentional here?

StevenWuzz:
> find a way to compress .jpg images #

Just wanted to mention here since you guys have probably taken CSCC73: The Huffman Algorithm could be used to compress such an image (and it is actually the best image compression algorithm AFAIK)

Ammar-T:
@StevenWuzz Haven't taken C73 yet unfortunately, but thanks for pointing that out. I'm gonna look forward to it!

src/app/utils/FabricUtils.ts
Ammar-T: Yeah, the `0` should be a `1` here, good catch.

StevenWuzz (approved):

JoelWiebe (approved):
Thanks @Ammar-T, looks great

Tested:
- Adding background image
- Replacing background image
- Removing background image
- jpg cannot be uploaded

l3n0ire (approved):

PR #140, opened 2022-01-26
Issue 128 nest boards within projects

### LINKED ISSUE #128: Nest boards within projects
In the main dashboard, add projects as the top-level that contain boards

For teachers:

- Projects created or subscribed to by the teacher are shown under “My Projects”
- Therefore the teacher will have both join and add icons available
- When creating a board, teachers have the option to select which project it belongs to

For students:

- As students cannot create projects or boards, only projects subscribed to by students are shown under My Projects
- Therefore students will have only the join icon available 

NOTE: Join codes will now operate on a per-project-level rather than on a board level

image

Teachers can create new projects from the main dashboard
image

l3n0ire:
## Details

- Nest Boards within projects
- Removed public projects and boards
- Will delete old boards from Firebase once approved

Closes #128


JoelWiebe (changes requested):
Amazing work! 

I just noticed some pretty minor things and provided a few suggestions.

Tested:
1. "Create board" option is enabled -> can we disable this option until after the first project is created?
2. Created first project -> (a) can we remove the pop-up dialogue box; (b) also I had to refresh the page for the project to be displayed
3. Clicked on project to open -> can we remove this pop-up dialogue box too
4. Created a new board -> (a) successfully defaulted the project to the current project, however, when creating the second board the project field was empty and I had to select the project; (b) when creating both boards there was an internal server error (500), but when I navigated back to the project, the new board was successfully created
5. Opened a board then clicked the return to dashboard button -> can we return to the current project rather than returning to the projects page?
6. Opened board configurations -> can we remove the following from the board configurations: (a) the public board option and (b) the members tab
7. Joined an existing project -> (a) can we remove the dialogue box; (b) had to refresh browser for project to show

All major functionality works really well. Students have appropriate restrictions in terms of creating projects and boards. Students can successfully join projects and have access to all boards within. Really wonderful work -- so fun exploring this new project structure.

JoelWiebe:
@l3n0ire Just out of curiosity, which user currently has ownership of the board: Use Cases for CK Board (that we plan to preserve) -- could we have this added to the new SCORE-dev project?

l3n0ire:
@JoelWiebe Good catch! Sorry about all the popups, I forgot to remove them after testing. I fixed the issues you've mentioned let me know if you find any more.

l3n0ire:
> @l3n0ire Just out of curiosity, which user currently has ownership of the board: Use Cases for CK Board (that we plan to preserve) -- could we have this added to the new SCORE-dev project?

The teacher that creates the board is the board owner. So not necessarily the project creator. Is this implementation ok or should only the project creator own all the boards?

JoelWiebe:
@l3n0ire Yes, I think that's a good way to go, where the teacher that creates the board is the board owner. 

What implications are there right now for being the board owner? I suppose this might come into relevance more once we get into managing board membership.

l3n0ire:
> @l3n0ire Yes, I think that's a good way to go, where the teacher that creates the board is the board owner.
> 
> What implications are there right now for being the board owner? I suppose this might come into relevance more once we get into managing board membership.

I don't think the board owner has any special permissions. Currently any teacher can change board permissions.

JoelWiebe:
@l3n0ire Awesome work getting these fixed!

I like the redirect inside the board once it's created!

It looks like just one of the issues is still occurring. I still need to refresh my browser after joining a new project (I login with student account and add the join code and submit; I remain on the project page with no changes; I refresh the browser, then the new project appears).

l3n0ire:
@JoelWiebe Fixed that issue. It should redirect to the project after clicking join.

JoelWiebe (approved):

JoelWiebe:
@l3n0ire Looks good to me!

@Ammar-T @StevenWuzz Feel free to review when you get a chance :)

src/app/app-routing.module.ts
```diff
@@ -5,13 +5,22 @@ import { DashboardComponent } from './components/dashboard/dashboard.component';
...
 
 const routes: Routes = [
-  { path: 'canvas/:boardID', component: CanvasComponent,
-    canActivate: [BoardGuard] 
+  {
+    path:'project/:projectID',
+    component:ProjectDashboardComponent,
+    canActivate:[ProjectGuard]
+  },
+  { path: 'canvas/:projectID/:boardID',
```
Ammar-T, line 20: Is it possible to change this to something like `project/:projectID/board/:boardID` ?

l3n0ire: @Ammar-T updated routing as per your suggestion

StevenWuzz (approved):

Issue #127, opened 2022-01-03
Create Workflow Management


JoelWiebe:
The teacher can view Workflow Management by clicking on the Workflow Tools icon in the top application navigation bar.

Workflow Management shows the status of all Bucket and Task Workflows and allows for editing of existing workflows. Task Workflows can also be shown or hidden by clicking the play/stop icons that add or remove these tasks from the student Workspaces.

Progress bars for tasks may show the fastest, slowest, and average progress of the class or a particular group (if assigned to a group). 

An example of the interface is shown below, populated based on existing Tasks and Buckets that were created using the Workflow Tools:

![image](https://user-images.githubusercontent.com/6416247/147904017-b3c94e96-3c26-4598-9bfc-319d2a16e85e.png)

JoelWiebe:
@Ammar-T Should we close this? I see the project status is set to done but the issue is still open.

PR #178, opened 2022-03-13
ISSUE-167: Distribution Workflow

### LINKED ISSUE #167: Distribute Posts Workflow
Randomize notes from a bucket or board to a set of buckets. 

For instance:
Source board/bucket = Ideas for Science Education (Board)
Action = Distribute: Randomly
Destination =
 + Group1 Bucket
 + Group2 Bucket 
 + Group3 Bucket
 - (not) Promising Ideas Bucket

Posts per bucket = 10 <- (this is a round-robin distribution where posts can be sent to multiple groups if all notes have been distributed once already)

Ammar-T:
### Small Fixes
- Fixed the toolbar image logo
- Added back the canvas settings button (I think these 2 issues stemmed from Jon's design PR)
- Remove extra and unnecessary console.log()
- Stop Bucket Modal's infinite loading spinner when a board has no buckets

### New Features
1. **Distribution Workflow** is for simply distributing posts from one source (Board or Bucket) into one or many destinations (Boards/Buckets/mix of both).
2. **Custom Criteria Workflow** is for more complex workflows which only move posts from one source to another destination depending on a specific criteria (e.g. # of likes, specified tags).

- As of right now, the UI for creating both workflows is available, however only the Distribution Workflows are able to be ran. The logic and functionality for custom criteria workflows will be done later.
- New tab called "Manage Workflows" can be used to view, run, and delete workflows.

Closes #167 #127


JoelWiebe (commented):
First of all, wow, awesome work! There's some complex stuff going on in there. 

Would you be open to changing the process of randomization a bit: Could we implement this as a round-robin where we skip a post if it already exists at the intended destination?

Essentially in pseudo-ish code:
sourcePosts //all posts from source
numSourcePosts //num posts in sourcePosts
distributedPostCount = 0 //counter for num distribute posts

for i=0; i<amountPerDest; i++
  for j=0; j<numDest; j++
    post = sourcePosts [distributedPostCount%(numSourcePosts-1)] //you could catch the case where numSourcePosts is 0 or 1 before the loop
    destPosts = get all posts from destination j
    if post is not in destPosts, then 
      add post to destination j
    distributedPostCount++ //increment to the next index, even if the post already exists in the destination and wasn't inserted (as it is already accounted for) 

---    
By any chance, are the sourcePosts sorted by recency? If so, this above approach may make for a more clear depiction of the distribution behaviour and an even distribution of older and more recent posts across all destinations. What do you think?

Ammar-T:
> First of all, wow, awesome work! There's some complex stuff going on in there.
> 
> Would you be open to changing the process of randomization a bit: Could we implement this as a round-robin where we skip a post if it already exists at the intended destination?
> 
> Essentially in pseudo-ish code: sourcePosts //all posts from source numSourcePosts //num posts in sourcePosts distributedPostCount = 0 //counter for num distribute posts
> 
> for i=0; i<amountPerDest; i++ for j=0; j<numDest; j++ post = sourcePosts [distributedPostCount%(numSourcePosts-1)] //you could catch the case where numSourcePosts is 0 or 1 before the loop destPosts = get all posts from destination j if post is not in destPosts, then add post to destination j distributedPostCount++ //increment to the next index, even if the post already exists in the destination and wasn't inserted (as it is already accounted for)
> 
> By any chance, are the sourcePosts sorted by recency? If so, this above approach may make for a more clear depiction of the distribution behaviour and an even distribution of older and more recent posts across all destinations. What do you think?

My current implementation actually already does this haha. When distributing posts, I do take into account posts that have already been distributed into that specific board/bucket and ignore them. However, you may have noticed this behaviour isn't always evident when distributing posts into boards (only boards, not buckets) and there's a good reason why:

When distributing posts into a board, we clone this post but assign it a new postID, then send it over to the destination board.
When distributing posts into a bucket, we add the post's ID into an array of post IDs which are stored in each bucket.

So, to know if a post already exists in a bucket, we simply loop over the post IDs in that bucket's array of post IDs. If it's found, then we know it's already in there and ignore the post for now.

To know if a post already exists in a board, well.... we can't. Since we assign each post a new ID before distributing it, there's no way for us to check if some post has already been sent to X board (since they won't have matching post ID), which is why sometime the same posts get distributed to boards. 

Ideally, if users want to avoid this, they should clear the bucket and then run the workflow again for now. Is that behaviour okay for now?

Ammar-T:
Good point on the order of distribution. I've updated it to follow your suggestion of shuffling the list of source posts first. This now makes the distribution truly random like it should.

I feel like allowing users to distribute posts into boards should still be an option. The copy workflow would be pretty repetitive I think, right? Not sure if it would be worth creating a new type for it. Perhaps we could add an info box where it states clearly how distribution works. It could mention the fact that "Distribution posts into boards could lead to duplicate copies..."

But, I'm good to discuss this further for a future PR!

JoelWiebe:
Brilliant, I love it!!

JoelWiebe (approved):

PR #180, opened 2022-03-14
Issue 164 background images broken

### LINKED ISSUE #164: Background images broken
Uploaded images often do not show completely. Perhaps there is an issue with the size of the image. 

Should we change our approach to compress and store the picture on the server rather than store the image as text as I believe we have done?

l3n0ire:
## Details
- need to run `npm install -force` since ngx-image-compress requires angular 13 but we're using angular 12. It still works on angular 12
- Images that are over 2mb are compressed to a size below 2mb. We have 1gb of firebase storage, so we can store around 500 images

Closes #164


src/app/components/add-board-modal/add-board-modal.component.ts
```diff
@@ -56,13 +61,36 @@ export class AddBoardModalComponent implements OnInit {
...
-  handleImageUpload(e) {
-    var file = e.target.files[0];
-    var reader = new FileReader();
-    reader.onload = (f) => {
-        this.bgImgURL = f.target?.result;
-    };
-    reader.readAsDataURL(file);
+  compressFile() {
+    // similar to code in configuration modal
+    // might refactor in future to avoid copy paste
```
Ammar-T, line 66: You commented it here and forgot perhaps, but I think having it refactored would be nice.

Ammar-T, line 66: You commented it here and forgot perhaps, but I think having it refactored would be nice.

JoelWiebe:
Great work! Thanks for allowing jpg again too. 

To reduce storage space, what do you think about creating two new functions (in another lower priority PR sometime in the future):
1) Somehow checking if an image being uploaded has already been uploaded and if so using that URL
2) When removing or changing an image, check to see whether the image is being referenced by any other board and, if not, delete the image from firebase

Let me know!

JoelWiebe (approved):
This looks good to me! I'll make another PR for looking into redundant or orphaned images. @Ammar-T if you're good with the changes, let's merge :)

Ammar-T (approved):

Issue #184, opened 2022-03-16
Add default message for task modal


Ammar-T:
When a board has no task created for it, opening the task modal is just an empty white box. 

We need to display some default message along the lines of "No task available." when there is no associated task with a board.

PR #187, opened 2022-03-16
ISSUE-172: Permanently show Task + move Task icon

### LINKED ISSUE #172: Make Task icon more obvious
Currently the task icon is hidden among other icons in the top navigation bar, perhaps we can move the Task description icon to the right of the title or somewhere more salient? (or maybe there can be a way to select to display/hide the contents across the top of the screen so they can read while they work)?

Ammar-T:
## Details

- Moves task icon to left side of toolbar
- When clicking the task button, it'll stick a snackbar to the top left with task info. If users want to read the full task, they can click "View Full Task".

#### Created a snackbar service queue
- This allows us to store state of all snackbars needed. Material only allows one snackbar at a time, which is why we need this service to handle multiple snackbars.
- e.g. task snackbar is showing, user wants to create post so the "Click anywhere on canvas to create post" snackbar appears. Now, we want to bring back our old task snackbar back since we didn't purposely close it, so this service handles all of that logic.

Closes #172 #184


JoelWiebe (commented):
Wow, quick implementation. You're flying!

Neat concept and good catch with the queue of snackbar items.

I notice we have two settings icons. One of those can probably be removed. Also what do you think about centering the snack bar, just so it's not eclipsing the board title? Additionally, do you think we should widen the snackbar max width and put the title and description on the same line to save space? Just an idea -- if possible!

Ammar-T:
@JoelWiebe So, I moved the snackbar for tasks to the bottom as well now. This should stop them from blocking anything important on the screen. I also made them a bit more wider and restricting the description text to be 2 lines max, which will reduce the height and maximize width. Let me know if these changes are good enough!

JoelWiebe (approved):
Yep, that'll do!

PR #188, opened 2022-03-17
ISSUE-175: Color-coded Tags

### LINKED ISSUE #175: Color code tags within each board
Currently we just store tags as strings, but we could create a new model for a tag that will be made up of the (1) tag name and (2) a colour representing that tag (with the option for a default colour, which could be a very light grey).

When a user opens a post and sees the list of tags associated with that post, the background colour of each tag should match the colour representing that tag.

Ammar-T:
## Details

- Adds the ability to choose a colour when creating a new tag in the "Create Board" modal and the canvas' "Settings" modal.
- Will assign a default colour if no colour specified.
- You will notice the colour coded tags in the post modal and the posts inside buckets/list.

Closes #175


JoelWiebe:
@Ammar-T Thanks for making quick work of this!

For some reason my board config icon disappeared -- is this something that happened on my end?
Screen Shot 2022-03-17 at 11 43 03 AM

Ammar-T:
@JoelWiebe Actually, it seems the settings button was added back already in a previous PR and is already in `develop`. So I've updated this branch to pull changes from `develop`.

You should be able to see it again now if you update this branch locally. (`git pull origin issue-175`)

Ammar-T:
Good call on the reset button, I'll add that. 

I'll also add the hardcoded tags now as well ("Idea" - light blue, "Question" - lighter orange, "Needs Attention" - red).

As for the behaviours for those respective tags, I think we should handle them in a separate issue?

JoelWiebe:
> Good call on the reset button, I'll add that.
> 
> I'll also add the hardcoded tags now as well ("Idea" - light blue, "Question" - lighter orange, "Needs Attention" - red).
> 
That would be great
> As for the behaviours for those respective tags, I think we should handle them in a separate issue?
Yes, that's perfect!

Ammar-T:
@JoelWiebe Added!

JoelWiebe:
Screen Shot 2022-03-17 at 4 24 24 PM
Any idea how my coloured tags options disappeared?

Ammar-T:
@JoelWiebe Not sure, are there any errors printed to the console maybe? I've checked my older and newer testing boards and they still have their tags.

Ammar-T:
@JoelWiebe Actually I just noticed that the screenshot you sent, the UI looks different compared to the one I just pushed.

Did you pull my new changes?

JoelWiebe (approved):
@Ammar-T Hahaha, I double and triple checked everything earlier, then when I logged back in I realized that I had the server open rather than localhost! Haha, now that I'm back on localhost, everything looks great :)

PR #201, opened 2022-03-19
ISSUE-189: Update viewport on pan swipe

### LINKED ISSUE #189: Move Posts Bugs
There are at least two bugs regarding the movement of posts:
1) Start with cursor on a post then pan away (using two fingers on a trackpad). You will notice that the four-arrowed cursor remains and indeed if you click on a blank spot on the canvas the post will open or if you click and drag the post will jump to your location
2) If you move a post, then move the cursor to a blank spot on the canvas and pan around in cirlces (with two fingers on a trackpad) for a few seconds, then you will no longer be able to click on or move the post you had just previously moved. To fix this, you can click the arrow keys on your keyboard which will reset the cursor and allow you to move the post.

Ammar-T:
## Details

- Previously, we did not properly update the viewport values since we did not call fabric's `setViewportTransform(vpt)`. 

Closes #189

### DIFF
--- a/src/app/components/canvas/canvas.component.ts
+++ b/src/app/components/canvas/canvas.component.ts
@@ -697,7 +697,7 @@ export class CanvasComponent implements OnInit, OnDestroy {
         if(!vpt) return;
         vpt[4] -= options.deltaX;
         vpt[5] -= options.deltaY;
-        this.canvas.requestRenderAll();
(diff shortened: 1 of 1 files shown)


JoelWiebe (approved):
Nice! That seems to do the trick.

PR #204, opened 2022-03-19
ISSUE-202: Fixed inaccurate dragging of posts

### LINKED ISSUE #202: Inaccurate dragging of posts
When clicking (with the intention to move the post) a post from a non-center area (near the title, bottom of the post, very left of the post, etc...), you may notice the post teleports so that the cursor is now placed in the centre of the post.

This is not intended. So, if someone were to drag a post by holding it from the top left, the cursor should stay on the top left area of the post while it is being moved around on the canvas.

@ me below if you were not able to understand this faulty behaviour and would like a video sample of the bug.

Ammar-T:
## Details

- When posts are being moved and updated, the positional update wasn't taking into account which specific area of the post was clicked. 

Closes #202

### DIFF
--- a/src/app/components/canvas/canvas.component.ts
+++ b/src/app/components/canvas/canvas.component.ts
@@ -593,8 +593,11 @@ export class CanvasComponent implements OnInit, OnDestroy {
       if (e.target) {
         var obj = e.target;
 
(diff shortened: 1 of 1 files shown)


JoelWiebe (approved):
Oooh, very clean feel moving notes now! I like it.

l3n0ire (approved):

PR #206, opened 2022-03-21
Issue 177 filter tags list view

### LINKED ISSUE #177: Filter by tags in List View modal
As we start off adding search and filtering functionality into our buckets and lists modals, we can start off with filtering by tags in just the list view.

l3n0ire:
## Details

- Filter by tags in list view
- Only posts that contain all of the filter tags will be displayed

Closes #177


JoelWiebe (commented):
@l3n0ire, amazing work! Thanks for making quick work of this.

If we wanted to denote that these filters are for tags, would it make sense to change the text to something like this?
image

JoelWiebe (approved):
@l3n0ire Thanks!

Ammar-T (approved):
👍

PR #208, opened 2022-03-22
ISSUE-163: Updated Canvas Synchronization

### LINKED ISSUE #163: Re-work updating posts' location in real-time
Currently, when posts are moved (for each x-coordinate/y-coordinate changed) we update Firebase with the new location. Although, you can imagine how many writes and reads this would create in a canvas with 10s of posts and 10s of users.

We want to update this method, so that:

1) When User1 clicks on a post with the intention to move it, lock the post for all other users and add a red border to the post (indicating that it's locked and being moved eventually)
2) User1 moves the post around and drops it at the destination location. 
3) Using Fabric Animations, animate the post to move from its initial destination to the destination for all other users.
4) Unlock the post and remove the red border indicating that the move is now finished.

Ammar-T:
## Details

- This PR implements a whole new way of updating and synchronizing posts/canvas-events between multiple users connected to a board.

### Post Movement
- When User A first holds down a post with the intent to move it to a new location, User B will receive that event and will lock that post on their canvas (also add a red border). This is to prevent multiple users from moving the same post concurrently.
- Once User A moves the post to its next location, User B will see the post being animated to its final location and once finished animating, it will unlock the post and remove the red border.

 
Closes #163


JoelWiebe:
@Ammar-T Awesome work, I'll take a look at this soon. Would we be able to swap the red border with a grey border (to draw less attention to "stuck" notes and reserve red borders for "Needs Attention" posts)?

Ammar-T:
Updated to grey!

Ammar-T:
@JoelWiebe Yeah so the reason the notes don't lock/move when the same user opens a board on 2 different devices is because of the synchronization logic. 

I only lock the post for all users who are not moving the post. Since the same user is going to be connected from 2 devices, it will never lock for the other device, since it's technically the same user. If we didn't handle this, the post would end up locking for the person who is about to move it. Obviously this doesn't work because it won't allow the mover to even move it haha.

To counter this, I think when a user logs in from one device A and then tries to log in from another device B. It should log them out of device A. Ideally, we shouldn't have the user connected to boards from different devices. And I think this is a fine limitation to have. Should we break this out into a separate issue?

----

Yeah I like the idea of changing the posts fill rather than just the border. I also noticed it was a little hard to notice when only the border changes, especially when zoomed out a little. On top of the fill colour, I can also try to maybe change the opacity of the post.

Ammar-T:
@JoelWiebe I should've worded that better, instead of the same user being on 2 different devices, we would check if the same user is logged in through multiple sessions. So, if the same user tried to login through a second tab, it would automatically log them out of their first one. This would also still hold if instead of 2 tabs, it were 2 different devices.

JoelWiebe:
@Ammar-T Oh yes, haha, that is a wonderful solution! I didn't even know that was a possibility :)

PR #214, opened 2022-03-22
ISSUE-212: Fix post background colour

### LINKED ISSUE #212: Fix post background colour
New posts mysteriously no longer have a background colour, then after the browser is refreshed the post background colours of new posts go black
Screen Shot 2022-03-22 at 9 01 56 AM

breity:
## Details
- Correctly sets the post background color on creation.

Closes #212


JoelWiebe (approved):
Thanks, you're a life saver!

Ammar-T:
I realized that we didn't set colour to the post when creating it from the bucket modal, so fixed that in the latest commit. 

@JoelWiebe I don't think you need to re-test but if you want you could.

Ammar-T (approved):

PR #221, opened 2022-03-23
Feat/trace

### LINKED ISSUE #19: Export posts to csv
For research, we need to be able to log user events (i.e., moving posts, tagging, commenting, etc.) and export this data as a CSV file including the post, comment, and tag contents so that researchers can analyze student work and activities.

SCORE has a similar export feature. Perhaps we can discuss this in more detail.

StevenWuzz:
## Details

- [CK-board export CSV template](https://docs.google.com/spreadsheets/d/1pMqKYt-hc7AQ4kvP2Ey4pQCTNvijqAFN/edit#gid=671197637)
- Tracing for the eventTypes listed in evenTypes page in the google sheets
- Button in project dashboard toolbar to download tracing data as csv

Closes #19


src/app/services/canvas.service.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+    
+    likeCanvasPostServer(postId: string, userId: string, boardId: string): void {
+        this.likesService.add({
+            likeID: Date.now() + '-' + userId,
+            likerID: userId,
+            postID: postId,
+            boardID: boardId
+        }).then(() => this.tracingService.traceVotedPostServer(postId, 1));
+    }
```
Ammar-T, line 118:
```suggestion
    likeCanvasPost(postId: string, userId: string, boardId: string): void {
        this.tracingService.traceVotedPostClient();
        this.likesService.add({
            likeID: Date.now() + '-' + userId,
            likerID: userId,
            postID: postId,
            boardID: boardId
        }).then(() => this.tracingService.traceVotedPostServer(postId, 1));
    }
```
Why can't we do a single function instead of having 2 separate ones? This would decrease some of the complexity.

StevenWuzz: That's what I thought before too..but I think the separation has some advantages such as:

1. Clear separation between what's done on the client side, and what's done on the server side
2. If there's a bug, it's much easier to catch which one causes. For example, I was able to locate the "double-tracing" incident (which, as we discussed, was intentional for now) caused by deleting a post really quickly. It would have been slightly harder if they were combined into one

src/app/components/post-modal/post-modal.component.ts
```diff
@@ -228,33 +209,35 @@ export class PostModalComponent {
...
       
     if (this.isLiked) {
-      this.likesService.remove(this.isLiked.likeID).then(() => {
-        this.isLiked = null
-        this.likes = this.likes.filter(like => like.likerID != this.user.id)
-      })
+      const postId = this.isLiked.postID;
+      const likeId = this.isLiked.likeID;
+
+      [this.likes, this.isLiked] = await this.canvasService
```
StevenWuzz, line 227: @Ammar-T To continue the previous discussion, I've also found that some parts (such as this one) require the objects to be returned to the caller (otherwise it would cause a bug, but maybe there is an alternative solution?). If we were to combine into one function, that means the client update would be done only after the call to the firebase is completed. However, based on the original code, I don't think that's an intended behaviour

src/app/services/canvas.service.ts
Ammar-T: I still think splitting it into 2 functions is excessive since we're kind've forcing the whole client vs server thing across the entire codebase. 

It should already be clear if we're calling 1 function then both client and server tracing will happen within that 1 function. Having a client and server function for each action just seems a little too much in my opinion.

StevenWuzz: @Ammar-T how about the second point?

src/app/components/post-modal/post-modal.component.ts
Ammar-T: Based on the original code, the client update only happens after the firebase call finishes. Look at the `.then(....`

src/app/services/canvas.service.ts
Ammar-T: I think we'll just sorta have to deal with it lol. You're right it would make it easier to debug, but not sure if duplicating all functions would be worth it.

StevenWuzz: @l3n0ire what do you think about this?

l3n0ire: I think its better to keep it one function. Like Ammar mentioned, we'd have to make similar changes across the whole code base. You can debug it by putting a break point after the client code for ex.

l3n0ire:
Not really relevant to the PR ,but @Ammar-T @StevenWuzz  what should we do with package-lock.json? 

I was thinking of using `npm install` to install new packages only and commit the updated package-lock.json. For building locally or on heroku we can install the node modules using `npm ci`  which doesn't modify package-lock.json.

Ammar-T:
Good point Colin, I think we all may be using different node versions which might be messing with the package lock files.

Do you think we should all first upgrade to the most stable node version and then proceed with your idea (committing a fresh package lock)?

l3n0ire:
Yeah sure. I installed node version 17.8.0 and npm version  8.5.5. Should I commit the fresh package-lock to master. I'll update the documentation too.

Ammar-T:
@l3n0ire yeah sounds good, also commit to develop as well I think?

JoelWiebe:
@StevenWuzz Thanks for making the updates to the code -- I see you're tracking reading already too, that's awesome!!

What do you think about adding a traceActivity column that identifies the type of activity that is being performed (e.g., create post, modify post, move post, read post, upvote post, create comment, add tag, remove tag, etc.)? I am also happy to add this as a future issue since this is not required, but may be a nice-to-have :)

Also, did you have any luck catching the case where a board within the project has no posts, it looks like I'm still getting an error?

src/app/interfaces/trace.ts
```diff
@@ -0,0 +1,27 @@
...
+    commentModifiedTextCounter: number
+    postModifiedUpvote: number
+    postTagNameAdded: string[]
+    postTagNameRemoved: string
+    postModifiedLocationX: number | null
+    postModifiedLocationY: number | null
+    postDeleted: number
+    bucketId: string
+    bucketName: string
+    postRead: number
```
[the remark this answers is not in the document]
StevenWuzz: Won't that imply the Trace collection will have different set of fields for each document?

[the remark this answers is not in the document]
Ammar-T: Maybe something like 
```
Trace {
    serverTimestamp
    clientTimestamp
    ... any other mandatory fields

    eventType: EventType
    event: Event
}

interface Event {
   ... any common fields between all events
}

LikeEvent implements Event {
    likeAmount 
    .. any other trace data for a like event
}
```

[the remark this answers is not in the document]
Ammar-T: With the above example, no it wont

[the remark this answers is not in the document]
StevenWuzz: @Ammar-T I honestly still don't get it..does it mean we will have a different collection called Event then?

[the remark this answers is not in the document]
Ammar-T: No event isn't it's own collection it's just a data structure used inside trace which is a collection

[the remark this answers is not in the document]
StevenWuzz: So the trace contains another collection?

JoelWiebe (changes requested):
First of all, wow, this is a lot of work. Sorry, we probably should have broken this up a bit to make it easier and get smaller bits of feedback at a time. Great work though, this is a lot to build. 

Tested
- read
- create post
- modify title and message
- modify title
- Modify message
- create comment
- create upvote
- remove upvote (from canvas)
- tag post
- remove tag from post
- move post
- delete post
- move post to bucket

NOTES:
- Can you add the serverTimestamp when a post is read, currently it displays -1 in that column
- I notice that the trace does not distinguish between whether the title or the message/description for a post has changed, but this makes sense and should be just fine (the modified counter goes up if either were modified)
- For comments, can you also add the postId so that we know which post is being commented on
- For removing an upvote, would it be possible to mark that as a -1; alternatively, later when there is a column that identified the trace activity that column could state "removed upvote"; currently the upvote column is a 0 so there is nothing to identify what changed in that trace row
- Deleting a post appears to delete two posts according to the trace data (with different post IDs) but the same title and message; how are we handling deletions, are there multiple copies of a note or is there only one copy and we are essentially hiding a post from the canvas, listview, and all buckets (which would be ideal for if we have a trashcan and ever want to revive a note)
- Are we still planning to add a traceActivity column that logs the action (e.g., "read", "upvote", "remove upvote", "create post", "modify post", etc.) ... if so, I can create the full list of these labels
- Are we still able to reorder the columns to match the template

Let me know what is feasible, and we can always turn these into additional features to help spread and share the weight of this work!

src/app/interfaces/trace.ts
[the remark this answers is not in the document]
l3n0ire: @Ammar-T if we refactor trace to use eventType and event we would run into issues during the csv export. Since events have different fields, trace object would have different schemas (after the event object is flattened). In that case we would have to use a different library. [Json2csv ](https://www.npmjs.com/package/json-2-csv) supports different schemas and default values for fields. If you're good with this i'll start refactoring the code.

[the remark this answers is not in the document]
Ammar-T: @l3n0ire Oh I see what Steven meant now, do you think we should store the event object as a string then so that event doesn't have different fields? The string would just be created by doing `JSON.stringify(event)`. 

You can go with whatever seems best, I just wanted to throw this out there as well!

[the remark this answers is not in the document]
l3n0ire: Yeah thats pretty hacky lol. I'll try with eventType and event and json2csv and see how far I get.

JoelWiebe:
@l3n0ire Here is the CSV template for reference: https://docs.google.com/spreadsheets/d/1pMqKYt-hc7AQ4kvP2Ey4pQCTNvijqAFN/edit#gid=671197637

Thanks for all the refactoring and taking on this work. When you have a chance, could you take a look at some of those items I noticed in the change request as well?

@Ammar-T regarding deleting notes, what kind of implementation did we have (i.e., was there a redundant copy of a note created rather than having a single note created that is marked as hidden?) I had noticed some interesting behaviour as noted above in the change request

l3n0ire:
@JoelWiebe I added a page called "Events" to the google sheets. These are the class names used the action types (e.g., "read", "upvote", "remove upvote" ..). I can add these class names as a column in the csv export

l3n0ire:
@JoelWiebe All the events in the eventType page in the google sheets are ready, except for deleting posts. (Same events that steven did). I'll work with Ammar to see how posts are deleted. I might remove it from this PR and create a separate issue for tracing deleting posts.

l3n0ire:
Currently working on migrating tracing to express server

l3n0ire:
@encorelab/score-dev Ready for testing and review.
 [Testing document](https://docs.google.com/spreadsheets/d/1KS1-TX2aYhLV4QUpNqHc2kCqoF5S_gOQ-ilWNS9AXvM/edit#gid=1913009351)
[csv template](https://docs.google.com/spreadsheets/d/1pMqKYt-hc7AQ4kvP2Ey4pQCTNvijqAFN/edit#gid=619966165)
You can download the csv export with the download button on the project dashboard

JoelWiebe:
@l3n0ire, this is amazing! You always do such great work. Thanks, I see that columns are re-ordered and eventType has been added. 

I tested the features and have my logs of what I tested below.

Before merging:
- I notice that the logs have entries from a "test trace" board from a "Test express mongo" project even though I am downloading logs from another project; do you know how these logs got here? 
- Once we get these removed let's go ahead with the merge!

Future issues:
- The file currently has the name CK_Trace; could we modify this so that the file name includes project name as well as the date and time (e.g., "CK-Trace My-First-Project 2022-06-08 at 3.01.16PM"? This way when we download many files, we do not end up with duplicate copies of the same file, but rather timestamped, identifiable copy. I will create a separate issue for this.
- I also noticed that there is no event type and no logs for BUCKET_REMOVE_POST. This can be addressed later. Let's get the rest all merged in for now!

Tested
* Empty board produces no log entries
* Create post - good (create post event)
* Edited note - good (modified counter increased)
* Edited note again - good (modified counter increased)
* Upvote - good (value of 1)
* Remove upvote - good (value of -1)
* Add to bucket - good (proper bucket and post references)
* Remove from bucket - TODO (no logs)
* Post deleted - good (created delete event)
* Create comment - good (logs comment text and id)

l3n0ire:
@JoelWiebe entries from other projects no longer show up in the export. Thanks for taking the time to test!

backend/src/server.ts
```diff
@@ -42,6 +43,7 @@ app.use("/api/comments", isAuthenticated, comments);
 app.use("/api/notifications", isAuthenticated, notifications);
 app.use("/api/groups", isAuthenticated, groups);
 app.use("/api/auth", auth);
+app.use("/api/trace", trace);
```
Ammar-T, line 46:
```suggestion
app.use("/api/trace", isAuthenticated, trace);
```

backend/src/socket/events/bucket.events.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  static async handleEvent(
+    input: SocketPayload<BucketEventInput>
+  ): Promise<BucketEventInput> {
+    bucketTrace.movePostToBucket(input, this.type);
+    return input.eventData;
+  }
+
+  static async handleResult(io: Server, socket: Socket, result: number) {
+    // dont need to emit for now
```
Ammar-T, line 17: I think we should emit it either way even if nothing will be listening for it.

backend/src/socket/events/post.events.ts
```diff
@@ -167,6 +181,23 @@ class PostTagRemove {
...
+    await postTrace.read(input, this.type);
+    return input.eventData;
+  }
+
+  static async handleResult(
+    io: Server,
+    socket: Socket,
+    result: PostTagEventInput
+  ) {
+    // dont need emit this for now
```
Ammar-T, line 197: Same thing here, emit either way.

docs/Trace.md
```diff
@@ -0,0 +1,19 @@
+# Trace Documentation
+
+### What is trace?
+
+Trace logs all user activity on boards. This data is exported to csv.
+
+### How to create a new trace
```
Ammar-T, line 7: Thanks for adding this, it looks super useful

frontend/package.json
```diff
@@ -21,9 +21,12 @@
     "@angular/platform-browser": "~12.0.1",
     "@angular/platform-browser-dynamic": "~12.0.1",
     "@angular/router": "~12.0.1",
+    "export-to-csv": "^0.2.1",
+    "express": "^4.17.1",
```
Ammar-T, line 25: What do we need express for in the frontend?

frontend/src/app/components/post-modal/post-modal.component.ts
```diff
@@ -79,6 +79,7 @@ export class PostModalComponent {
     public fabricUtils: FabricUtils,
     @Inject(MAT_DIALOG_DATA) public data: any
   ) {
+    //console.log(data);
```
Ammar-T, line 82:
```suggestion
```

Ammar-T:
@l3n0ire Really neat work! Just a couple small things here and there but should be good to merge after that if all the test cases (manual ones) pass as well.

Ammar-T (approved):

PR #223, opened 2022-03-23
ISSUE-117: Post modal three dot menu issue

### LINKED ISSUE #117: Display issue with three vertical dot menu
image

When post editing permissions are turned on and someone opens a peers' post, the entire three vertical dot menu should be hidden (like it is when post editing permissions are turned off). 

However, the three vertical dot menu is shown but empty when opening a peers' post.

Ammar-T:
## Details

- Before, we had a 3-dot vertical menu with options to edit and delete the post inside the post modal.
- I scrapped the 3-dot menu and just listed the options as buttons since it doesn't make sense to have a whole menu for just 2 buttons
- Also added a confirm dialog when user wants to delete the post

NOTE: You can test it out on staging instead of having to pull locally now. Check the URL I sent in our slack channel.

Closes #117

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -38,6 +38,7 @@ import { BucketsModalComponent } from './components/buckets-modal/buckets-modal.
 import { ListModalComponent } from './components/list-modal/list-modal.component';
 import { ToolbarComponent } from './components/toolbar/toolbar.component';
 import { SnackBarComponent } from './components/snackbar/snackbar.component';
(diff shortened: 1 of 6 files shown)


JoelWiebe (approved):
Good thinking, it increases visibility of these items!

Just curious, do notes become permanently deleted or are they still stored in firebase? Maybe in the future we can implement a "trash can" if a teacher needs to recover some accidentally deleted posts.

Ammar-T:
No, they won't be permanently deleted, so yeah we can still access them if needed in the future.

Issue #199, opened 2022-03-19
Update top nav bar with blue


JoelWiebe:
Can you perhaps send me the palette tool with the colours you are using or a couple of blue options to choose from?

I do like the purple, but want to stay a little closer to the traditional Encore blue if possible :)

PR #224, opened 2022-03-24
ISSUE-210: Fix snackbar CSS/design

### LINKED ISSUE #210: Fix snackbar CSS/design
Update the snackbar component to complement the new CK-Board theme. 

Choose valid colours, font-sizes, etc. to match the new theme.

breity:
## Details
- Restores snackbar styling and background color.
- Improves visibility of snackbar buttons by removing accent color. Accent color doesn't provide enough contrast with the dark background. Colored buttons do look nice, so in the future we can revisit. Maybe we can think about this more in conjunction with #199.
- Adds a little space between snackbar buttons.

Closes #210

### DIFF
--- a/src/app/components/snackbar/snackbar.component.html
+++ b/src/app/components/snackbar/snackbar.component.html
@@ -1,11 +1,11 @@
-<div *ngIf="!description" style="display: flex; flex-direction: row;">
+<div *ngIf="!description" class="actions">
   <div class="title" [innerHTML]="title">
     {{ title }}
   </div>
-  <button mat-stroked-button color="accent" (click)="close()">
+  <button mat-button (click)="close()">
     Close!
   </button>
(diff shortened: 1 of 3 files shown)


JoelWiebe (approved):
Looks great! Yes, we can think about re-adding colour to the buttons when doing re-colouring later. 

@Ammar-T I notice that the snackbars for creating groups and workflows aren't showing, is this also a CSS thing? Maybe we can open another issue for those as they aren't necessary for Sunday's run.

Ammar-T:
@JoelWiebe I don't think we had a snackbar when creating a workflow. But you're right, I'm not sure where the snackbar for creating buckets went haha. We can look into it after the run.

Ammar-T (approved):
👍

Ammar-T:
@JoelWiebe Oh wait no, those snackbars were introduced in #209 which isn't part of this branch yet.

PR #225, opened 2022-03-24
ISSUE-197: Resize buckets/list-view posts

### LINKED ISSUE #197: Resize buckets/list-view posts
For buckets and list-view (available from the nav tools at the top of the canvas)
- Make posts wider (similar to ratio on canvas)
- Consider making pop-up full-screen width
- Columns may responsively expand to (perhaps) 4 columns

breity:
## Details
- Increases buckets and post list dialog max width.
- Adds a close button to dialogs.
- Adds responsiveness to post lists and increases html post width.
- Cleans up html post styles.
- Adds keyboard accessibility to html posts.

Closes #197.

### DIFF
--- a/src/app/components/buckets-modal/buckets-modal.component.html
+++ b/src/app/components/buckets-modal/buckets-modal.component.html
@@ -1,4 +1,9 @@
-<h1 mat-dialog-title>CK Buckets</h1>
+<h1 mat-dialog-title>
+  <span>CK Buckets</span>
+  <button mat-icon-button [mat-dialog-close] aria-label="Close">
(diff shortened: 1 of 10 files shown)


JoelWiebe (approved):
Booyah!  Responsiveness at 1, 2, and 3 columns looks much better. Good idea adding that "x" to close as well. I'm good with this for the run. It doesn't have to be today, but what do you think about making the max-width of the modal wider for four columns so the posts at 4 columns can be a bit wider too?

Ammar-T (approved):
👍

PR #247, opened 2022-04-03
Issue 213 notification system

### LINKED ISSUE #213: Add User Post Subscriptions and Notifications
Create Board Notifications tool for receiving notifications, for instance, upon:
- Comment, upvote, or tag by another user to their posts

Board Notifications could include a Bell icon in the top navigation bar with a numerical counter indicating the number of notifications. Once clicked, a tray could be displayed that shows a list of notifications. Selecting a notification could open the modal pop-up for that post. 

SCORE has something like this:
Screen Shot 2022-03-22 at 9 07 22 PM

l3n0ire:
## Details
- User will be notified whenever a user comments, tags or likes their posts
  - no option to turn on/off notifications yet (should I add this?)
- When a user clicks on a notification, the post modal opens and the notification is marked as viewed
- For now notifications are deleted after they have been viewed
- Mark all as read will mark all notifications as read

Closes #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ This Common Knowledge (CK) Board is the latest rebuild of the ENCORE Lab's colla
 
 #### 1. Install Node.js and Angular
 * Node: https://nodejs.org/
+  * Use the LTS versions. v16 works, v17 has some compatabily issues 
 * Angular: ```npm install -g @angular/cli ```
 
 #### 2. Clone the repository


src/app/components/notification-dropdown/notification-dropdown.component.ts
```diff
@@ -0,0 +1,100 @@
...
+import { NotificationService } from 'src/app/services/notification.service';
+import { PostService } from 'src/app/services/post.service';
+import { PostModalComponent } from '../post-modal/post-modal.component';
+
+@Component({
+  selector: 'app-notification-dropdown',
+  templateUrl: './notification-dropdown.component.html',
+  styleUrls: ['./notification-dropdown.component.scss']
+})
+export class NotificationDropdownComponent implements OnInit {
```
Ammar-T, line 15: Need to also implement `OnDestroy` if you want to use `ngOnDestroy()`.

src/app/services/notification.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  observable(userID: string, handleAdd?: Function, handleModification?:Function, handleDelete?:Function) {
+    return this.notificationCollection.ref.where("userID", "==", userID).onSnapshot((snapshot) => {
+        snapshot.docChanges().forEach((change) => {
+          const doc = change.doc.data();
+            if (handleAdd !== undefined && change.type === "added") {
+                handleAdd(doc)
+            }
+           else if (handleModification !== undefined && change.type === "modified") {
```
Ammar-T, line 28: Will we ever need to listen for updates to a notification? I feel notifications should just be a one-time message and shouldn't be editable.

I'm guessing you added this for the case of a notification being "marked as read"? We wouldn't need to listen for that right, since only the user who marked it as read would be able to mark it as read.  So, instead we could just update the UI directly after they click "Mark all as read" instead of listening for it.

Ammar-T:
I noticed we're creating notifications inside the other services (e.g. `postService`). In the future, if we want to disable notifications, we would have to somehow pass this info to the individual services as well then. 

Could we separate the logic of creating notifications and whatever the individual services are doing? For example, if we tag a post then the notification shouldn't be created inside the `postService` but maybe in `canvas.component.ts` after the `update()` finishes? 

I also talked to Steven about creating some middleware service (`canvas.service.ts`) for the canvas component where the notification could be created perhaps? You could check out his PR. The main purpose of this middleware service would then be tracing, actually updating the post (or whatever the action is), then creating a notification if needed.

src/app/models/notification.ts
```diff
@@ -0,0 +1,21 @@
...
+    return {
+        notificationID :uuidv4(),
+        text :text,
+        timestamp :timestamp,
+        viewed  :viewed,
+        userID :userID,
+        postID  :postID,
+    }
+
+}
```
Ammar-T, line 20:
```suggestion
export class Notification {
    notificationID:string;
    text:string;
    timestamp:number;
    viewed:boolean;
    userID:string;
    postID:string;
 
    constructor(userID: string, text: string = "", postID: string="", timestamp: number = Date.now(), viewed: boolean = false) {
        this.notificationID = uuidv4();
        this.userID = userID;
        this.text = text;
        this.postID = postID;
        this.timestamp = timestamp;
        this.viewed = viewed;
   }
 }```

src/app/services/notification.service.ts
l3n0ire: Good point. Should we also have the same behavior for deleting notifications? I assume only the user themselves would be able to delete notifications.

Ammar-T: Yeah, same behaviour for deletes makes sense too.

l3n0ire:
@Ammar-T @StevenWuzz I refactored notifications to use the canvas.service.ts from steven's pr. I commented out all the tracing stuff. Hopefully it wont be too difficult to merge later on.

Ammar-T:
@l3n0ire Thanks for doing this. Instead of commenting the tracing stuff out, do you think it would be easier to just completely remove it to avoid conflicts with Steven's PR. 

Also, I feel there still may be a lot of conflicts since we're eventually gonna consolidate the duplicate calls to canvas.service. (Doing `addPost` instead of `addPostClient` and `addPostServer`).

To make it easier for you, I think it would be fine if instead of pulling Stevens changes, you just create a new canvas.service in this PR and only add the necessary functions for this PR. This would avoid all the extra commented code and then less merge conflicts in the future as well.

PR #248, opened 2022-04-06
Issue 246 create post in list view

### LINKED ISSUE #246: Create post in list view
- Add "create post" button to list view
- By default will only be visible to user with "download" button to place on board
- similar to create post in buckets

l3n0ire:
## Details
-  Option to add posts in list view

Closes #246

### DIFF
--- a/src/app/components/canvas/canvas.component.ts
+++ b/src/app/components/canvas/canvas.component.ts
@@ -151,6 +151,9 @@ export class CanvasComponent implements OnInit, OnDestroy {
       autoFocus: false,
       data: {
         board: this.board,
+        user:this.user,
+        centerX: this.canvas.getCenter().left,
+        centerY: this.canvas.getCenter().top
       },
     });
   }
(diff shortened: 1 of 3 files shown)


l3n0ire:
I noticed that bucket only posts are showing up in list view. Currently bucket only posts and lists view only posts are created the same (post with empty fabric object). Maybe we can differentiate them by adding an additional field to posts like `postType: "list"`?

Ammar-T:
@l3n0ire I think it's okay to have bucket posts show in the list view. The list view is supposed to have pretty much every post no matter the type. The bucket only posts should be fine to appear in the list view but shouldn't show on the canvas.

JoelWiebe:
@l3n0ire Great question! Yes, I like @Ammar-T 's suggestion. Conceptually, we can think of list view as holding all the posts (with and without a location), the canvas holding only posts that have a location, and buckets holding assigned posts (with or without a location).

JoelWiebe (approved):
Tested:
- Creating posts from list view
- Adding posts from list view to canvas

This feature is really neat. It's still so cool to me to click the download icon and have them drop onto the canvas :)

Ammar-T (approved):

PR #249, opened 2022-04-09
ISSUE-243: Make likes from canvas asynchronous

### LINKED ISSUE #243: Num likes and comments is incorrect
The number of likes and comments displayed on the canvas does not necessarily match the number in the modal post pop-up. 

For example: project/1647961013141-YxXrM7L8g4WHlgvlmn81hrTtbEy1/board/1648228593348-YxXrM7L8g4WHlgvlmn81hrTtbEy1

image

If you work on this, I can provide the join code for your convenience :)

Ammar-T:
## Details

- Each time the like button is clicked, instead of incrementing/decrementing the value by 1, we check the db for total likes first. This is to stop race conditions which cause the like amounts to be inaccurate.
- Also made liking posts from canvas async which fixes the issue of clicking the like button twice really quickly causing the number to go up by 2. Obviously, this should never happen since one user can only like a post once.

Closes #243

### DIFF
--- a/src/app/components/canvas/canvas.component.ts
+++ b/src/app/components/canvas/canvas.component.ts
@@ -418,10 +418,10 @@ export class CanvasComponent implements OnInit, OnDestroy {
         existing = this.fabricUtils.updatePostTitleDesc(existing, obj.title, obj.desc);
         this.canvas.requestRenderAll()
       } else if (event == CanvasPostEvent.LIKE) {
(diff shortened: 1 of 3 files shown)


l3n0ire (approved):

JoelWiebe (approved):
Tested:
- upvote and remove downvote from canvas
- upvote and remove downvote from modal pop-up

Changing vote value even fixed a post that previously had an incorrect value

PR #254, opened 2022-04-29
Issue 228 removed public board from board configuration

### LINKED ISSUE #228: Remove public board option
Public Board is still a toggle option in the Board Configurations. Can we remove this toggle and all code related to this feature as I believe we no longer plan to support public boards?

l3n0ire:
## Details

- Removed public board from board configuration

Closes #228


Ammar-T:
@l3n0ire Should we also remove the `public` field from the `Board` model?

l3n0ire:
> @l3n0ire Should we also remove the `public` field from the `Board` model?

Yep, good catch. We also need to remove it from the board guards.

Ammar-T (approved):
👍

PR #255, opened 2022-05-03
Issue 234 set default zoom

### LINKED ISSUE #234: Set Initialization Zoom Depth for Boards
To avoid users being lost in a large background layout, it would be helpful if teachers could specify in the board configurations the zoom depth that the board opens to (i.e., 30% so that users don't immediately need to zoom out when they visit a particular board)

Screen Shot 2022-03-24 at 10 10 45 PM

l3n0ire:
## Details

- Field to set initial zoom in board configuration
- Fixed bugs with background image upload
- When the background image is updated, the initial zoom is reset to 100%

Closes #234

### DIFF
--- a/src/app/components/configuration-modal/configuration-modal.component.html
+++ b/src/app/components/configuration-modal/configuration-modal.component.html
@@ -14,9 +14,9 @@
       </mat-slide-toggle>
       <mat-form-field appearance="outline" style="width:100%; margin-top: 20px;">
         <mat-label>Initial Zoom</mat-label>
-        <input matInput [(ngModel)]="initialZoom" #initialZoomInput="ngModel" type="number" required>
+        <input matInput [(ngModel)]="initialZoom" #initialZoomInput="ngModel" type="number" min="1" required>
         <mat-error *ngIf="initialZoomInput.invalid">
-          Initial Zoom is required.
+          Please enter a value greater than 0
         </mat-error>
       </mat-form-field>
       <div class="image-form">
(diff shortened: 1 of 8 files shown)


src/app/components/configuration-modal/configuration-modal.component.html
```diff
@@ -4,7 +4,17 @@ <h1 mat-dialog-title>Board Configuration</h1>
...
         <mat-label>Board Name</mat-label>
-        <input matInput [(ngModel)]="boardName">
+        <input matInput [(ngModel)]="boardName" #boardNameInput="ngModel" required>
+        <mat-error *ngIf="boardNameInput.invalid">
+          Board Name is required.
+        </mat-error>
+      </mat-form-field>
+      <mat-form-field appearance="outline" style="width:100%; margin-top: 20px;">
+        <mat-label>Initial Zoom</mat-label>
+        <input matInput [(ngModel)]="initialZoom" #initialZoomInput="ngModel" type="number" min="1" required>
```
Ammar-T, line 14: Can we use a slider here instead of number input? I think it would be more user-friendly and we can probably find one from Material-UI.

Ammar-T:
Code-wise looks good, just the slider addition and everything else looks fine. Thanks for adding docs to some of the image handling functions as well.

JoelWiebe (commented):
This is wicked awesome!!

Just one small idea, once the initial zoom value is set, I notice that the board zoom is updated to match the initial zoom value. Can we also centre the x and y position of the canvas (the same as if the user just refreshed the browser page to be centred on the canvas)? This way they could have the exact view that people have when entering the page.

Really great work! (I like your work on the slider too!)

JoelWiebe (approved):
Love it!

PR #263, opened 2022-05-18
Make read task snackbar toggleable

### LINKED ISSUE #230: Close Tasks by Re-clicking Task Icon
Rather than needing to click "close" in the snackbar to close a task, can we re-click the task icon in the top left to close the task snackbar? That way if someone clicks once to read they can quickly re-click again to close without moving their cursor across the screen... sometimes its the little things :)

DCH25:
## Details

- Read Task icon on the board now only queues a snackbar if there's none currently open 

Closes #230

### DIFF
--- a/src/app/components/canvas/canvas.component.ts
+++ b/src/app/components/canvas/canvas.component.ts
@@ -427,6 +427,11 @@ export class CanvasComponent implements OnInit, OnDestroy {
     const title = this.board.task.title ? this.board.task.title : 'No task created!';
     const message = this.board.task.message;
 
+    if (this.snackbarService.snackBar._openedSnackBarRef) {
+      this.snackbarService.dequeueSnackbar();
+      return;
+    }
+
     const openDialogCloseSnack = () => {
       this.dialog.open(TaskModalComponent, {
         width: '500px',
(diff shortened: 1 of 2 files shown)


l3n0ire (approved):

src/app/components/canvas/canvas.component.ts
```diff
@@ -427,6 +427,11 @@ export class CanvasComponent implements OnInit, OnDestroy {
     const title = this.board.task.title ? this.board.task.title : 'No task created!';
     const message = this.board.task.message;
 
+    if (this.snackbarService.snackBar._openedSnackBarRef) {
```
Ammar-T, line 430: Could we make a function inside snackbarService to see if there's a currently-opened snack bar instead? 

We should keep the 'snackBar' field private and inaccessible from other classes. I forgot to do this initially.

DCH25: makes sense, just added this change

src/app/services/snackbar.service.ts
```diff
@@ -63,6 +63,10 @@ export class SnackbarService implements OnDestroy {
     );
   }
 
+  snackbarIsOpen(): Boolean {
+    return Boolean(this.snackBar._openedSnackBarRef);
+  }
```
Ammar-T, line 68: Thanks for adding this! 

I don't think you need to convert it to 'Boolean' here - it's fine to just return a 'boolean'.

Ammar-T (approved):
Just one more small suggestion then feel free to merge!

PR #266, opened 2022-05-19
Task title now required with task description

### LINKED ISSUE #233: Require a task title
If a task exists, it must have a title. The description is optional.

Currently the system allows a task description without a task title, but then the default title message is that no task exists:
Screen Shot 2022-05-04 at 9 58 04 AM

LunarFang416:
## Details

- If user enters a Task description, they are now required to enter a valid title
- If the title is just whitespace ("    "), you still will not be able to submit, regardless of whether title description is entered
- "test" and "&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; test" are examples of valid titles which user can update or create when initially creating the board.

Closes #233


src/app/components/add-board-modal/add-board-modal.component.html
```diff
@@ -87,5 +87,5 @@ <h4>Set Background Image:</h4>
 </div>
 <div mat-dialog-actions>
   <button mat-button (click)="onNoClick()">Cancel</button>
-  <button mat-button  cdkFocusInitial [disabled]="boardNameInput.invalid || projectInput.invalid" (click) ="handleDialogSubmit()">Create Board!</button>
+  <button mat-button  cdkFocusInitial [disabled]="(taskMessage && !taskTitle.trim()) || (taskTitle && !taskTitle.trim()) || boardNameInput.invalid || projectInput.invalid " (click) ="handleDialogSubmit()">Create Board!</button>
```
l3n0ire, line 90: taskMessage is optional so !taskTitle.trim() should be enough. Nice that you added in the trim. I don't think the other fields do that so we should prob add that in the future and refactor how we do form validation.
```suggestion
  <button mat-button  cdkFocusInitial [disabled]="!taskTitle.trim() || boardNameInput.invalid || projectInput.invalid " (click) ="handleDialogSubmit()">Create Board!</button>
```

LunarFang416: If we just do !taskTitle.trim(), wouldn't that mean you need to have a task title at all times to update/create the board? 
The reason i added the trim was because if the user enters just whitespaces (and i don't check for whitespaces), when you try to see the view the task on the page, it would be blank. So by doing !taskTitle.trim(), it sees if the title was whitespace, if it was it wont let it update.

l3n0ire: in `add-board-modal.component.ts` taskTitle is defaulted to `''`. Task title is mandatory I believe, so you need to have it all times. You reasoning makes sense, just that taskMessage is optional so we don't have to check.

LunarFang416: oh ok, in the ss for the issue it showed 'No Task Created' so I thought having a task (also task title) was not mandatory at all times.

l3n0ire: @JoelWiebe are tasks mandatory for boards? If thats the case then no change needed

JoelWiebe: @l3n0ire @LunarFang416 Tasks are not mandatory at this point; however, if a task message exists then a task title is required :)

JoelWiebe: This does bring up an interesting point though, if tasks are not required should we disable the task icon rather than have the student open it up just to see a message that no task was created, haha?

l3n0ire (approved):
mb i misunderstood how the tasks worked

src/app/components/add-board-modal/add-board-modal.component.html
JoelWiebe: Sorry for the scope creep, but what do you think about us also removing the "View Full Task" button in the snackbar if the task can already fully fit in the snackbar?

JoelWiebe: And finally... I see that in at least some of the branches, tasks are mandatory when creating a new board. I'll create a follow up PR to remove tasks as a requirement when creating a new board (to make board creation super quick and simple)!

JoelWiebe:
@l3n0ire You were actually correct! But now we're changing it :) https://github.com/encorelab/ck-board/issues/267

JoelWiebe:
@LunarFang416 just a quick addition -- could we add this title warning (similar to what we have in the Add Post tool shown below) where a red warning appears when the description receives focus when the title is blank? Sorry I missed this previously!

Screen Shot 2022-05-26 at 1 41 04 AM

LunarFang416:
@JoelWiebe Just added the error message. Shows for both board creation and board config updates.

Screen Shot 2022-05-26 at 12 30 28 PM

Screen Shot 2022-05-26 at 12 30 03 PM

Ammar-T (approved):

JoelWiebe:
@LunarFang416 Thanks for your great work on this! I just noticed a small bug. Now when I create a new board without a task I get this error in the backend:
Screen Shot 2022-06-08 at 10 18 30 AM
However, when I create a new board with a task I get no such error. Can you please look into this?

LunarFang416:
@JoelWiebe I fixed the changes and the big doesn't occur now.

Ammar-T:
@LunarFang416 Feel free to merge if everything's good here.

PR #268, opened 2022-05-25
Text is hyphenated and also wraps around line in posts

### LINKED ISSUE #173: Canvas text wraps mid-word
Text in the title and body of posts on the canvas wrap to a new line on any character, rather than wrapping whole words or at appropriate hyphenation points. Can text-wrap options be set for these items on the canvas?

Screen Shot 2022-03-09 at 10 21 27 AM

LunarFang416:
## Details

- Text now doesn't cut off in the title and description and is hyphenated and moved over to the next line
- I did a lot of testing to make sure it works properly, but I would appreciate if you could do some quick checks as well so that we know it works how we want it too

A few screenshots:

Screen Shot 2022-05-25 at 1 45 05 PM

Screen Shot 2022-05-25 at 1 45 33 PM

Closes #173

### DIFF
--- a/src/app/components/fabric-post/fabric-post.component.ts
+++ b/src/app/components/fabric-post/fabric-post.component.ts
@@ -1,108 +1,202 @@
 import { Component, Inject } from '@angular/core';
 import { fabric } from 'fabric';
-import { POST_DEFAULT_BORDER, POST_DEFAULT_BORDER_THICKNESS } from 'src/app/utils/constants';
+import {
+  POST_DEFAULT_BORDER,
+  POST_DEFAULT_BORDER_THICKNESS,
+} from 'src/app/utils/constants';
 
-const AUTHOR_OFFSET = 65
-const DESC_OFFSET = 80
-const CONTENT_EXTRA_HEIGHT = 55
+const AUTHOR_OFFSET = 65;
+const DESC_OFFSET = 80;
+const CONTENT_EXTRA_HEIGHT = 55;
 
 @Component({
   selector: 'app-fabric-post',
   templateUrl: './fabric-post.component.html',
-  styleUrls: ['./fabric-post.component.scss']
+  styleUrls: ['./fabric-post.component.scss'],
 })
 export class FabricPostComponent extends fabric.Group {
+  constructor(@Inject(Object) options: any) {
+    var canvas = new fabric.Canvas('c');
+    // FROM : http://jsfiddle.net/illumine/Avvxn/
 
-  constructor(@Inject(Object) options:any) { 
-    var title = new fabric.Textbox(options.title, {
-      name: 'title',
-      width: 280,
-      left: 18,
-      top: 60,
-      fontSize: 18,
-      fontWeight: 'bold',
-      fontFamily: 'Helvetica',
-      fill: '#000000',
-      splitByGrapheme: true
-    });
-  
-    var author = new fabric.Textbox(options.author, {
-      name: 'author',
-      width: 300,
-      left: 18,
-      top: title.getScaledHeight() + AUTHOR_OFFSET,
-      fontSize: 13,
-      fontFamily: 'Helvetica',
-      fill: '#555555',
-      splitByGrapheme: true,
-    });
+    function wrapCanvasText(t, canvas, maxW, maxH) {
+      if (typeof maxH === 'undefined') {
+        maxH = 0;
+      }
+      var words = t.split(' ');
+      var formatted = '';
+      var context = canvas.getContext('2d');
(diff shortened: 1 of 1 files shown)


src/app/components/fabric-post/fabric-post.component.ts
```diff
@@ -1,117 +1,211 @@
...
-      name: 'author',
-      width: 300,
-      left: 18,
-      top: title.getScaledHeight() + AUTHOR_OFFSET,
-      fontSize: 13,
-      fontFamily: 'Helvetica',
-      fill: '#555555',
-      splitByGrapheme: true,
-    });
+    function wrapCanvasText(t, canvas, maxW, maxH) {
```
Ammar-T, line 22: Wrapper function looks good! Do you mind moving it into `utils/FabricUtils.ts`? That file is responsible for any helper functions associated with anything fabric-related. 

You can also remove the `canvas` parameter from this function as well since `_canvas` is already set in `utils/FabricUtils.ts`.

Ammar-T (changes requested):

JoelWiebe:
@LunarFang416 Amazing, thanks for making quick work of this. I'm totally good with this functionality but for some reason the hyphen isn't displaying on long words -- am I doing something wrong:
Screen Shot 2022-05-26 at 1 33 14 AM

LunarFang416:
Thats very strange, I did the same and this is what i get:
Screen Shot 2022-05-26 at 11 00 43 AM

It doesn't add the text-wrap to the existing posts (posts before i implemented the feature), but it does it to all new posts.

DCH25:
@LunarFang416 @JoelWiebe I think the context.measureText() method might be rounded or isn't completely accurate for some letters. Ex. the 's' and 'r' in the image. I'm pretty sure this is what's causing the inconsistency but I'm not sure what the fix would be.

image

LunarFang416:
I'll look into this more, and see what I can do. Thanks for catching this!

DCH25:
Another small thing to add is just to apply word wrap after editing a post, but otherwise LGTM!

src/app/components/fabric-post/fabric-post.component.ts
```diff
@@ -1,117 +1,157 @@
...
   selector: 'app-fabric-post',
   templateUrl: './fabric-post.component.html',
-  styleUrls: ['./fabric-post.component.scss']
+  styleUrls: ['./fabric-post.component.scss'],
 })
 export class FabricPostComponent extends fabric.Group {
+  constructor(
+    @Inject(Object) options: any,
+    ) {
+    var fabricUtils: FabricUtils = new FabricUtils();
```
Ammar-T, line 22: Instead of instantiating it here, you should inject it through the constructor. 

See 'canvas.component.ts' constructor for example.

Ammar-T (changes requested):

src/app/components/fabric-post/fabric-post.component.ts
LunarFang416: @Ammar-T I was trying to inject the fabricUtils before like this as well: 
```js
constructor(
    @Inject(Object) options: any,
    protected fabricUtils: FabricUtils,
  )
```
But this was giving and error because we we're exporting the FabricPostComponent and using it in other classes and when we initialize the FabricPostComponent class it gives the following error: `An argument for 'fabricUtils' was not provided.` I assume this is because we do not give it the fabricUtils in the constructor. Would I have to pass the fabricUtils when initializing the `FabricPostComponent`. 
I was unsure about that since I didn't see something like that in the code so that is why I did it the other way.
Maybe I'm doing it wrong. What would be the correct way of doing this? Thanks.

Ammar-T: I see what you mean. I think leave it how you have it for now since that works. 

We'll need to redo how we create Fabric posts anyways. It doesn't make sense to have a whole angular component for it since we're just doing all the logic within the constructor lol. 

Once we change that, we'll find a nicer way of accessing fabric utils rather than forcing it through the constructor which doesn't make a lot of sense.

JoelWiebe:
@LunarFang416 How is it that something that looks so simple very rarely is, haha. We can come back to this as it is a lower priority. 

I was having some trouble building this now though, I'm running npm install then ng serve but it cannot find a bunch of modules. Is this just something on my end?

Ammar-T:
@LunarFang416 Is this ready to merge or what's the status on this?

LunarFang416:
@Ammar-T i've been working on this for the past few weeks and couldn't make any progress; the same bugs are still happening. I just put it on hold now xd.

PR #275, opened 2022-05-30
groups api and endpoints configured (#269)

### LINKED ISSUE #269: Groupings - Model & API Endpoints
Inside `/backend`, we need to create a `Group` model that will store all data associated with a group.

```
class Group {
    groupID: string // unique id for group
    name: string // group name
    members: string[] // array of user ids of all members in group
    ... other data as you may seem fit
}
```

We also need to create API endpoints to `create`, `update`, `read`, and `delete` groups. 

You should synch with @DCH25 regarding whatever endpoints he may need to create his UI.

LunarFang416:
## Details

- Implemented the API and endpoints for the grouping feature of the ck-board
- Will begin testing backend with postman

Closes #269

### DIFF
--- a/backend/src/api/groups.ts
+++ b/backend/src/api/groups.ts
@@ -53,18 +53,19 @@ router.post('/:id', async (req, res) => {
     res.status(200).json(updatedGroup);
 })
 
-router.delete("/:id/users", async (req, res) => {
+router.post("/:id/users/add", async (req, res) => {
     const { id } = req.params;
+    const { users } = req.body;
 
-    const updatedGroup = await dalGroup.removeUsers(id);
+    const updatedGroup = await dalGroup.addUser(id, users);
     res.status(200).json(updatedGroup);
-})
+});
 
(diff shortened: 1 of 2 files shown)


Ammar-T:
@LunarFang416 Thanks for the quick work! 

I was wondering if you could add 1 more endpoint for adding users to a group. Similar to how you already have one for removing a user. You can rename the remove user route to `/groups/:id/users/remove` and then add this `/groups/:id/users/add` as well. You can use this as a [reference](https://github.com/encorelab/ck-board/blob/develop/backend/src/repository/dalBucket.ts#L61).

Also, would appreciate if instead of only removing a single user for that endpoint, we can remove any amount of users. Would also expect the same behaviour for adding users (i.e. allow adding multiple users in bulk rather than only one).

Let me know if that makes sense!

LunarFang416:
@Ammar-T Thanks!. I just made the changes.

Ammar-T (approved):
Feel free to merge if the postman tests were ok.

PR #282, opened 2022-06-08
Detailed error messages

LunarFang416:
## Details

- Error messages now outputted in the try catch blocks in the backend are more detailed 
- Previously error messages were simply `throw new Error("500")` which did not give info on the specific error, however now the error caught in the catch block is returned
- This will help future debugging if backend problems occur. 

An example output of an error message:
```json
[ERROR] Error: {
 "errors": {
  "boardID": {
   "name": "ValidatorError",
   "message": "Path `boardID` is required.",
   "properties": {
    "message": "Path `boardID` is required.",
    "type": "required",
    "path": "boardID",
    "value": ""
   },
   "kind": "required",
   "path": "boardID",
   "value": ""
  }
 },
 "_message": "BoardModel validation failed",
 "name": "ValidationError",
 "message": "BoardModel validation failed: boardID: Path `boardID` is required."
}

```

### DIFF
--- a/backend/src/repository/dalBoard.ts
+++ b/backend/src/repository/dalBoard.ts
@@ -5,7 +5,7 @@ export const getById = async (id: string) => {
     const board = await Board.findOne({ boardID: id });
     return board;
   } catch (err) {
-    throw new Error("500");
+    throw new Error(JSON.stringify(err, null, " "));
   }
 };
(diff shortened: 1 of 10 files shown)


Ammar-T (approved):
Thanks for adding this

PR #284, opened 2022-06-11
Stop storing fabric representation of post in DB

Ammar-T:
## The Problem

Previously, for each post we were storing the fabric object that represents that post on the canvas. (the field `fabricObject` inside the `Post` model). However, this has grown to be highly inefficient since we realistically should not be storing such large objects even if they're stringified.

Another problem caused by this method of storage is that when someone wants to make changes to a post from outside the canvas page (e.g. the new separate CK Workspace page), we can't update the fabric object since there's no active canvas. Often times, if we want to update a post then we have to find it's fabric representation from the canvas (through `fabricUtils.ts --> getObjectFromId()`), although, if there's no canvas then this becomes really difficult.

## The Solution
Now, instead of storing an entire stringified fabric object inside each post, we'll only store the important parts for it's fabric representation (i.e. the post border, color, border thickness, opacity, etc.).
```
- fabricObject: "{key: value, key: value, key: value, ... 100s more unnecessary key-value pairs that we don't care about}"
+ displayAttributes: { position, lock, borderColor, borderWidth, fillColor, opacity }
```

Although, this also requires us to change how we handle certain events like liking, tagging, moving posts. Instead of having to update the `fabricObject` and then updating the DB with it as well, we simply just update the specific fields that need updating (if any).


l3n0ire:
@Ammar-T could you also update move() in post.trace.ts after trace is merged?

Ammar-T:
> @Ammar-T could you also update move() in post.trace.ts after trace is merged?

Yep sure

PR #288, opened 2022-06-15
Issue 191 resize canvas

### LINKED ISSUE #191: Resize canvas on window resize
When the user resizes their window, the canvas currently does not resize to fit the new dimensions.

We need to listen for resizing events and resize this canvas accordingly. We also noticed if a user leaves the canvas page, resizes their window, and goes back the the canvas page, the canvas does not fit to the new dimensions. It still adapts to the previous dimensions.

DCH25:
## Details

- Canvas resizes with the window
- Right now board posts will scale down as much as the window does. A minimum could be added but I'm not sure if it's necessary since it only affects smaller resolutions, would like some thoughts on this. 

Closes #191


JoelWiebe:
Great work Daniel! Nice to have this feature at last.

If I understand your question correctly, I expect we don't need to explicitly specify a minimum width because the browser will have a minimum width right? As long as we can scale down to the minimum browser width we should be good. 

I do notice though that the nav bar throws off scaling at small widths. For instance if we have a long title that keeps the nav bar wide, then the nav bar and the canvas will be far wider than the browser width. Perhaps we can handle this as another issue or address this by shortening board title by reducing the title length and adding an ellipsis (...) at narrower widths. 

Should we commit this and create a separate issue for title resizing? @Ammar-T, thoughts?

Screen Shot 2022-06-16 at 11 20 01 AM

Screen Shot 2022-06-16 at 11 16 44 AM

Ammar-T:
@JoelWiebe I think we can fix the title overflow issue in this PR as well as I think it's an easy fix. 

@DCH25 I think you could use the text-overflow css property.

DCH25:
I added the fix for the title scaling to the PR. 

@JoelWiebe 
In regards to the minimum post width something like the image below would be pretty unreadable at small window sizes. The user could click the zoom button several times to read a post but it seems a little inconvenient. So I was thinking either set a minimum post size or maybe make scrolling a shortcut for zooming in and out.

image

JoelWiebe:
@LunarFang416 @zarifmahmud Do you mind doing a review (i.e., feel free to offer in-line comments, summative feedback, or simply approve if all looks good)?

JoelWiebe:
@DCH25 Oh interesting, what browser and OS are you using? You shouldn't need to hold control. The laptops we tested with touch pads did this automatically (supports two finger swipe for panning and pinching for zoom).

JoelWiebe (approved):

DCH25:
@JoelWiebe I'm on Windows 11 and the latest chrome version. Unfortunately I don't have a laptop to test pinching right now.

image

frontend/src/app/components/canvas/canvas.component.ts
```diff
@@ -396,6 +396,29 @@ export class CanvasComponent implements OnInit, OnDestroy {
...
 
+  onResize(event) {
+    let scaleX = event.target.innerWidth / this.canvas.getWidth();
+    // Without toolbar height
+    let scaleY = (event.target.innerHeight - 64) / this.canvas.getHeight();
+    let objects = this.canvas.getObjects();
+
+    // Resize all objects inside the canvas
+    for (var i in objects) {
+      objects[i].scaleX = (objects[i].getObjectScaling().scaleX) * scaleY;
```
Ammar-T, line 407: Is this a typo? Should it be `* scaleX` instead?

DCH25: I made it scale with height only to preserve the post's aspect ratio. This only really matters if the user decides to shrink the window on one axis only.

Ammar-T (approved):

LunarFang416 (approved):

PR #291, opened 2022-06-20
Issue 290 tracing permissions

l3n0ire:
## Details

- Added tracing permission to allow/disable tracing
- Tracing is disabled by default
- Added docs for "areas of improvement" for tracing

Closes #209


backend/src/socket/trace/base.trace.ts
```diff
@@ -17,6 +22,10 @@ export const createTrace = async (
   ) {
     throw new Error("Invalid board project or user in Trace Context");
   } else {
+    // check tracing enabled
+    if (!board.permissions.allowTracing) {
```
Ammar-T, line 26: Instead of querying the board each time, can we update `TraceContext` to have a boolean field called `enabled`? 

This way we can just do `if (input.trace.enabled) await createTrace(...)` inside each of the tracing events.

JoelWiebe:
@l3n0ire Oh sorry, ignore my other comment. I'll review this shortly.

Ammar-T (approved):
Looks good, feel free to merge after adding Joel's suggestions.

l3n0ire:
@Ammar-T  I noticed for the board events triggered by the configuration modal, the sockets are being emitted twice. Any idea what's causing this?

backend/package.json
```diff
@@ -18,8 +18,8 @@
     "build": "tsc",
     "dev": "ts-node-dev --respawn ./src/server.ts",
     "style:linter": "eslint . --fix --ext .ts",
-    "style:formatter": "prettier --check '**/*.{js,jsx,ts,tsx}'",
-    "style": "prettier --write '**/*.{js,jsx,ts,tsx}'"
+    "style:formatter": "prettier --check \"**/*.{js,jsx,ts,tsx}\"",
+    "style": "prettier --write \"**/*.{js,jsx,ts,tsx}\""
```
l3n0ire, line 22: single quotes don't work on windows.

l3n0ire:
@encorelab/score-dev how do you guys feel about adding git hooks? Basically it runs linters before each commit and prevents commits if the linter fails. https://prettier.io/docs/en/precommit.html

Ammar-T:
@l3n0ire Hmm which events specifically, is there a way to reproduce it I can play around with it later today.

l3n0ire:
> @l3n0ire Hmm which events specifically, is there a way to reproduce it I can play around with it later today.
@Ammar-T 
So in the tracing-permission branch, I went to a board and click on the update configuration button. On the backend I console log all the events in the` _listenForEvents` function and I get this. I see the same behavior on the develop branch
image

Ammar-T:
@l3n0ire hmm I'm wondering if there's maybe multiple connections opened then? Like from the client side we're connecting to the websockets twice? Inside the socket manager can you check how many users are connected (there should be 1 since it's just you locally).

l3n0ire:
> @l3n0ire hmm I'm wondering if there's maybe multiple connections opened then? Like from the client side we're connecting to the websockets twice? Inside the socket manager can you check how many users are connected (there should be 1 since it's just you locally).
Socket manager is showing that there's only one connection. Weird thing is that all the other events are emitted once, just the events triggered by update board configuration are emitted twice.

l3n0ire:
@Ammar-T fixed it
`const events = [
  ...postEvents,
  ...boardEvents,
  ...workflowEvents,
  ...notificationEvents,
  ...boardEvents,
  ...bucketEvents,
];`
Turns out we imported board events twice lmao

Ammar-T:
Oh my god lol it must've been a merge conflict gone wrong

PR #292, opened 2022-06-20
Issue 252 - Delete Boards

### LINKED ISSUE #252: Delete boards
Delete boards within a project which also removes any data associated with it (i.e. posts, likes, comments, etc.). 

The delete button can be somewhere inside the settings modal.

LunarFang416:
## Details
- Buckets can now be deleted from the setting modal.
- This deletes all entities related with a board, including Posts, Likes, Comments , Buckets , Notifications, Tags, Trace, Workflow, ProjectBoardData
- I also made some changes in the router guard so that routes to boards which don't exist (or ones that are deleted) return an error
- Older boards will give an error when deleting since the modal for the boards updated and now include the projectID, which was not present before and is needed for deleting the board

Screen Shot 2022-06-20 at 3 08 06 PM
With a confirmation dialog:
Screen Shot 2022-06-20 at 3 08 39 PM

Closes #252


Ammar-T:
@LunarFang416 Can you move the "Delete Board" button inside the modal's content area (not where the "Close", "Update" buttons are)? I think a warning-themed "mat-raised-button" under the "Upload new background..." button would be fine.

backend/src/api/boards.ts
```diff
@@ -52,4 +62,39 @@ router.get("/users/:id", async (req, res) => {
...
+      deletedBoard,
+      deletedPosts,
+      deletedLikes,
+      deletedComments,
+      deletedTags,
+      deletedTrace,
+      deletedWorkflow,
+      deletedBuckets,
+      deletedNotifications,
+      deletedProjectBoard,
```
Ammar-T, line 92: Don't think we need to be returning all of the deleted content as well, that'll be a lot of unwanted data being sent back.

backend/src/api/boards.ts
```diff
@@ -52,4 +62,39 @@ router.get("/users/:id", async (req, res) => {
...
+    const deletedPosts = await dalPost.removeByBoard(id)
+    const deletedLikes = await dalLike.removeByBoard(id)
+    const deletedComments = await dalComment.removeByBoard(id)
+    const deletedBuckets = await dalBucket.removeByBoard(id)
+    const deletedNotifications = await dalNotification.removeByBoard(id)
+    const deletedTags = await dalTag.removeByBoard(id)
+    const deletedTrace = await dalTrace.removeByBoard(id)
+    const deletedWorkflow = await dalWorkflow.removeByBoard(id)
+    const deletedProjectBoard = await dalProject.removeBoard(
+      deletedBoard!.projectID,
```
Ammar-T, line 79: Also, when you move all of this logic into one `dalBoard` function, create a session transaction so the deletes are consistent. Example here: https://github.com/encorelab/ck-board/blob/develop/backend/src/repository/dalPost.ts#L59

frontend/src/app/guards/project.guard.ts
```diff
@@ -54,6 +65,11 @@ export class ProjectGuard implements CanActivate {
     return this.project !== null;
   }
 
+  async isValidBoard(boardID: string) {
```
Ammar-T, line 68: Thanks for adding this check

LunarFang416:
Hey @Ammar-T, is this how you wanted the button to be?

Screen Shot 2022-06-22 at 9 31 25 AM

Ammar-T:
@LunarFang416 Could you instead make it look like the button above instead? So with the rounded corners and an icon before the text (trash icon).

I think it'll look nicer if they both match a bit. Thanks!

LunarFang416:
yep, that would look way better

JoelWiebe:
@LunarFang416 thanks for all your work on this!

I notice in one case, I deleted a board and that board and the ones created after it in the project view become named "picture_in_picture". Refreshing the browser does not fix this. However, I could not recreate this when I created a new project. Maybe this is the result of an old board or one that contains unexpected data?

For example, I was testing out the new dot-voting tool that Ammar was working on. Will this delete tool need to be updated every time a new type of data associated with the board (i.e., dot-voting) is added or removed?

Screen Shot 2022-06-23 at 12 29 05 PM

LunarFang416:
@JoelWiebe The reason that you're getting picture_in_picture for old boards (boards before this delete feature), is because before the Board Model did not have a projectID field. To delete a board completely, we need to delete the board from the project model as well since it contains an array of all the ids of the boards. 

So for older boards, the boards and everything related to it are being deleted, but their ids from the project model are not. So it still displays the deleted boards, but when it tries to get the board with that id from the backend, it returns null, because its been deleted. Thats why you get picture_in_picture, since the board is null. 

Also for future data associated with the board we will have to update the board delete functionality. Actually, I will also have to add the deletion of dot voting and remove the like deleting since it wont be used.

JoelWiebe:
Thanks @LunarFang416 for the info!

In that case, this looks good from my perspective. Once @Ammar-T signs off, we should be good to go. Also @l3n0ire feel free to review if you have time!

We'll just need to keep an eye on this when adding/removing features. Good to know :)

Ammar-T (approved):

JoelWiebe:
@LunarFang416 Looks like we are good to resolve conflicts, merge, then delete the branch. If all still looks good, feel free to go ahead :)

LunarFang416:
@JoelWiebe @Ammar-T Just finished integrating the deletion with upvotes, if that looks good, we can merge the branch.

backend/src/repository/dalLike.ts
```diff
@@ -0,0 +1,65 @@
+import Like, { LikeModel } from '../models/Like';
```
Ammar-T, line 1: This file was deleted in my upvotes PR. It should be dalVote now.

backend/src/repository/dalBoard.ts
```diff
@@ -47,12 +56,36 @@ export const update = async (id: string, board: Partial<BoardModel>) => {
...
 };
 
+export const remove = async (id: string) => {
+  const session = await mongoose.startSession();
+  session.startTransaction();
+  try {
+    const deletedBoard = await Board.findOneAndDelete({ boardID: id });
+    if (deletedBoard) {
+      await dalPost.removeByBoard(id);
+      await dalComment.removeByBoard(id);
```
Ammar-T, line 66: I think you need to add dalVote.removeByBoard here as well to remove all the upvotes for this board.

Ammar-T (changes requested):

backend/src/repository/dalBoard.ts
LunarFang416: oh ye i forgot to push the final changes xd. ill do it rn.

Ammar-T (approved):

PR #293, opened 2022-06-21
Issue 244 - Delete comments

### LINKED ISSUE #244: Delete Comment
Allow students to edit or delete their own comments; prevent students from editing or deleting others' notes. Allow teachers to delete anyone's comment.

"Deleted" comments should remain stored in the database, but tagged as deleted and hidden in its post

The delete "filled trash" icon may be presented in the top right corner of the comment (as displayed today), preferably, only on mouse hover of that comment (if relatively easy to implement). Tentatively, we can also display a confirmation dialogue (e.g., "Are you sure you want to delete this comment?") -- perhaps this dialogue could be removed in the future if we have some means of users recovering deleted comments.

Screen Shot 2022-05-26 at 1 08 42 AM

zarifmahmud:
## Details

- Teachers can delete anyone's comments.
- Students can only delete their own.
- Deleted comments immediately disappear for the one deleting, and comment counts are also updated.

Teacher view:

encore_teacher_comment

Student view:
encore_student_comment

Delete confirmation:

encore_comment_confirmation

Closes #244


Ammar-T:
Can you move the delete icon to the end of the row instead of it being next to the comment text? Also would prefer it to be vertically in-line with the commenter's name instead of being in-line with the comment text.

frontend/src/app/services/canvas.service.ts
```diff
@@ -105,6 +105,15 @@ export class CanvasService {
     }
   }
 
+  async deleteComment(commentID: string, postID: string){
```
Ammar-T, line 108: Shouldn't we be sending out the `COMMENT_REMOVE` event through socket service here?

frontend/src/app/components/post-modal/post-modal.component.ts
```diff
@@ -239,6 +239,25 @@ export class PostModalComponent {
...
   }
 
+  async deleteComment(comment: Comment) {
+    this.dialog.open(ConfirmModalComponent, {
+      width: '500px',
+      data: {
+        title: 'Confirmation',
+        message: 'Are you sure you want to delete this comment?',
+         handleConfirm: () => {
+          // this.socketService.emit(SocketEvent.COMMENT_DELETE, commentID);
```
Ammar-T, line 249: Remove comment please, I think you forgot to move this inside the `canvas.service.ts` function.

backend/src/api/comments.ts
```diff
@@ -23,4 +23,19 @@ router.get('/posts/:id', async (req, res) => {
...
 });
 
+router.delete("/:id", async (req, res) => {
+  const { id } = req.params;
+  const comment = await dalComment.getById(id)
+  let amount = -1
+  if (comment) {
+    amount = await dalComment.getAmountByPost(comment.postID);
+  }
+  const deletedComment = await dalComment.remove(id)
```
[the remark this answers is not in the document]
zarifmahmud: Should the catch throw an error, or just return a 404?

frontend/src/app/services/canvas.service.ts
zarifmahmud: Sorry I'm not exactly sure how the socket service works yet. What would the COMMENT_REMOVE event do?

backend/src/api/comments.ts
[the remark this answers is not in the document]
Ammar-T: Return this "res.status(500).end(error)"

JoelWiebe:
@zarifmahmud @Ammar-T What do we think about the trash icon only appearing beside a comment when we mouse over the comment (that way we aren't constantly looking at trash cans :)

frontend/src/app/services/canvas.service.ts
Ammar-T: It would essentially alert all users connected to the board that a comment was deleted, so they can update the comment count for the respective post on the canvas. You just have to add something like this `socketService.emit(SocketEvent.COMMENT_REMOVE, comment);`

zarifmahmud:
I've added the requested changes, let me know if there's anything else to add!

JoelWiebe (approved):
Nice, elegant solution :)

JoelWiebe:
@Ammar-T Does this look okay to you? Good to merge?

backend/src/socket/trace/post.trace.ts
```diff
@@ -138,6 +138,21 @@ const commentAdd = async (
...
+const commentRemove = async (
+  input: SocketPayload<CommentModel>,
+  eventType: string
+) => {
+  const trace = await createTrace(input.trace);
+  const comment = input.eventData;
+  trace.event = {
+    postID: comment.postID,
+    commentID: comment.commentID,
+    commentModifiedText: comment.comment,
```
Ammar-T, line 150: This shouldn't be a field for the comment remove event. It should be something similar to the event for removing a post, so just one field like "commentDeleted".

Ammar-T (changes requested):

backend/src/socket/trace/post.trace.ts
zarifmahmud: I kept the commentID field and added a commentDeleted to make it like the post remove event

Ammar-T (approved):

PR #305, opened 2022-06-29
Delete Buckets

### LINKED ISSUE #251: Delete buckets
Allow for the deletion of buckets. This might be possible by changing the Create Bucket tab under the Workflows tool to a Manage Buckets tab that includes options for deleting buckets.

LunarFang416:
## Details

- Buckets can now be deleted
- Downloading the bucket posts was giving an error, so i made a slight change in the backend to fix that
- I also found a bug where you couldn't download bucket posts from the List modal to the board, so i quickly added that as well
- UI for deleting buckets is like this:

Screen Shot 2022-06-29 at 3 37 06 PM

Closes #251


JoelWiebe:
Really sleek interface! I like the expanding delete delete section. For consistency should we add the number of buckets next to the arrow (like we do with comments)?
Screen Shot 2022-07-04 at 3 56 19 PM

Tags are not added or deleted from posts. (this works in the develop branch but not here; I tried this on the canvas). Is it possible that the backend changes are blocking tags from being added/removed?

LunarFang416:
@JoelWiebe I'm a bit confused. When we delete the buckets only the buckets are deleted but the posts remain. To my knowledge, tags aren't related with bucket deletion. Or is this a bug not related to bucket deletion?

LunarFang416:
I just tested it again, and it seems that this is probably an unrelated issue because I cant add tags to the posts either, regardless of whether they are on the board or in a bucket.

LunarFang416:
I'll try to see whats causing it, its probably some issues on my end.

LunarFang416:
@JoelWiebe fixed the bug where you couldn't add/remove tags to posts, hopefully it works correctly now.

JoelWiebe (approved):
New icons look good

For the bug, I tested
- Adding and removing tags
- Adding comments 
- Adding and removing votes

All seems to work now -- thanks!

backend/src/repository/dalPost.ts
```diff
@@ -58,12 +58,10 @@ export const removeByBoard = async (boardID: string) => {
 
 export const update = async (id: string, post: Partial<PostModel>) => {
   try {
-    const attrUpdate = formatAttributes(post);
```
Ammar-T, line 61: Why're we removing this? It will break on occasions where the post is updated (ie. add a "Needs Attention" tag, move the post around, refresh, and it'll disappear).

LunarFang416: This and the $set operator were actually causing an error when trying to download posts from the bucket view. Removing the $set didn't cause the error again, and it doesn't break when a post is updated.

Ammar-T (approved):

PR #307, opened 2022-06-30
Create workspace UI

Ammar-T:
## Details

- 

Closes #{ISSUE-NUMBER}


Ammar-T:
@JoelWiebe You can checkout this branch to see the workspace UI. Once you're on the canvas page, click on the "checkmark" icon button on the top toolbar (right side, most-left icon). 

Note: 
- no functionality. 
- clicking on post inside cards doesn't work yet. 
- you can drag cards left and right to swipe through them or just use "next", "previous" buttons

JoelWiebe:
Oh my goodness!! That card stack is slick. Very material feel to it.

How are you thinking that we could differentiate this for teachers and students? I.e.

- For the students, maybe we can hide the inactive task workflows AND show only the tasks relevant to them (assigned to either the whole-class or their group)
- For the teacher, do you think it would be useful for the teacher to be able to select from an option at the top to switch between views of different groups (i.e., select a group's name to see the task workflow view of that group -- allowing them to see the progress of the group and preview their views to ensure everything is setup correctly)?
- Where do you think the teacher's [Workflow Management tools](https://docs.google.com/presentation/d/1HCbXmwMkj3ZgOKnBdI348iymSdOGt4a3/edit#slide=id.g10b5d63c7a7_1_5) should exist? It seems that this Task Workflow view is closely related to Workflow Management. Do we create a button for the teacher toggle between this view (looking at how individual groups are progressing on their tasks) and management and watching and overview of all groups)?

How does it look if we give a little more screen real-estate to the opened task workflow on the right side (a little closer to half the width)?

Ammar-T:
@JoelWiebe Great points!

So I think we have to split up the views for teachers and students since there's not much overlap. 

For Students:
- We keep the current view as it is and this is going to be known as the "Student View"
- We only show them their active assigned workflows or active workflows assigned to the whole class
- We can also change the ratio of workflows to cards and split it half and half as you mentioned

For Teachers:
- We create a whole new view which will mostly include workflow management features and progress indicators on a per-group basis
- This will include some of the components from your mockup
- For now, I don't think we should allow teachers to view the perspective of students because that can get tricky. It'll be difficult to show them student views since student views will differ per group. It's a bit of added complexity that we can look into later.
- However, we can add some neat statistics to show their progress. Even though the teachers can't exactly see what the students see, they will still be able to thoroughly see their progress.

JoelWiebe:
@Ammar-T That's a great plan, let's do that. 

As you mention, maybe in the future we can allow the teacher to also select a group to open up that groups' exact Workspace view.

JoelWiebe:
@Ammar-T Did you have any more recent commits to push? I thought I had heard you had been working on this but I don't seem to see the latest update. Let me know! :)

Ammar-T:
@JoelWiebe Heads up, I added more commits since my last message to you so you will need to pull again.

I was playing around with the new workflows which sparked a few more questions in my mind and I'd love to hear your input.

Question:

For task workflows, when should the posts be transferred from the source board/bucket to the destination board/bucket? I realized that often times groups will be assigned the same posts, e.g. In Task Workflow 1, Group 1 gets Post X, Y, Z and Group 2 gets Post A, B, X, (Both were assigned Post X). Now we could do 2 things here:
- Each time a group finishes all their actions on their set of posts (e.g. they tagged each post once) and they "submit task", all of those posts will be sent to the destination. (Ex. Once Group 1 tags all their posts, X, Y, Z will be sent to destination, then Group 2 tags all their posts and A, B, X will be sent to the destination.)
- Once all groups finish their tasks, all of the posts will be sent to the destination. (Ex. Group 1 tags all their posts, Group 2 tags all their posts, then X, Y, Z, A, B posts will be sent at once to the destination).

I was leaning towards method 1, but there's a small development issue if we go that route. Once we send posts to other boards/buckets, we don't keep track of which ones are sent. So, if a post was already sent, it may be sent again if another group finishes their task on the same post. Ex. Group 1 tags post X, finishes their task and sends it to Board K. Group 2 tags post X, finishes their task and sends it once again to Board K. We could work around this with some extra work, but I wanted to see if you would even prefer the first method, so I don't put in the extra work for no reason haha.

JoelWiebe:
Brilliant questions! 

I'll discuss some ideas for the task workflow feature first, then get back to your questions.

Some ideas/suggestions:
1. Would we be able to have two classes of tasks (1) Generate (that has a destination only and (2) Review that has requirement options (a) tag, (b) comment, and upvote
2. Tag requirement when selected is that each post in the set should be tagged at least once
3. Comment requirement when selected is that each post in the set should be commented (given feedback) at least once
4. Voting requirement when selected is that each post can be voted on (i.e., I was consider having this would enable voting for the posts while disabling would hide voting for the posts -- is this easily implementable for the workspace, if so, maybe we could do something similar for tags and comments)
5. In terms of submission, I had been thinking about each post being able to be individually submitted, that way the teacher and students can see a sense of the progress of a task (with progress bars) and students can see their pile slowly disappearing
6. For post distribution, we are using the random distribution function to divide posts evenly among groups, right? Rather than manually selecting a number, could we automatically evenly distribute ALL posts from the source among all groups assigned (i.e., if one group, they get all posts; if two groups they each randomly get half of the posts; if three groups, they each randomly get a third of the posts)
7. Similarly, if we have a whole-class option, perhaps we could randomly split all of the posts in the source among each of the students (i.e., if there are 20 students, they each get 1/20th of the posts)

Typo:
- "Atleast one group required" -- missing a space

Questions:
1. I see there is (perhaps a stub) "ammar" group. Could we merge the grouping UI feature to see how that integrates?
2. Is the randomization function still intended provide duplicates. The last time I tried it, it just ran out of posts when it assigned all posts once and no students got duplicates. Maybe that is the best design for now -- duplicates might just make things confusion as your question points out and might not even make sense if we, for example, have two groups tagging the same post?
3. Are posts currently being removed from the source when they are added to the destination?
4. What do you think about splitting the "Manage Workflows" tab into "Manage Task Workflows" and "Manage Buckets Workflows" -- or within Manage Workflows having a way to toggle between viewing task and bucket workflows? 

Response to your questions:
1. For when to move the post to the destination, it should be fine to move the post to the destination immediately once a group/individual has submitted that post. I expect there may be cases where students don't finish their tasks and then posts would be stuck, unable to move to the destination. It should be fine if a post is attempted to be added to a bucket twice though, right, it just would only be added the first time? Also if we use the randomization and distribution methods discussed above, we could avoid multiple groups having the same post for the same task which may bypass these problems. 
2. For voting, perhaps we make this optional for now (since we only have dot-voting). Once up/down voting is implemented, perhaps we can require that every post in the set receives either an upvote or a downvote?

I do apologize if I was too distantly involved during the development of this feature and if these ideas require too many changes at this point. Perhaps we can meet to discuss these sometime soon. Also, I'm happy if a bunch of these are added as New Issues that others can help work on as well.

Ammar-T:
@JoelWiebe 

Regarding the ideas/suggestions, I like all of them! I'll try to get most of them in this PR, but the remaining ones I'll put into separate issues.

Thanks for catching the typo.

**Regarding the questions:** 

> I see there is (perhaps a stub) "ammar" group. Could we merge the grouping UI feature to see how that integrates?

Yep, I can do this.

> Is the randomization function still intended provide duplicates. The last time I tried it, it just ran out of posts when it assigned all posts once and no students got duplicates. Maybe that is the best design for now -- duplicates might just make things confusion as your question points out and might not even make sense if we, for example, have two groups tagging the same post?

No duplicates will appear as of the current implementation. If we're running out of posts, the last group will just get less posts or none at all if there's not enough to distribute.

> Are posts currently being removed from the source when they are added to the destination?

Not at the moment, I can change that as I've noticed Abubukker already has that implemented in his Workflows PR.

> What do you think about splitting the "Manage Workflows" tab into "Manage Task Workflows" and "Manage Buckets Workflows" -- or within Manage Workflows having a way to toggle between viewing task and bucket workflows?

100% agree. That tab is getting a little bloated and would be better if it was split up into separate sections.

Ammar-T:
@JoelWiebe 

**Things updated/added:**
- Merged in Groupings UI and is fully integrated
- Removed voting requirement for now (will add in separate issue)
- Updated comment requirement to only require 1 comment per post
- Updated tag requirement to only require 1 tag per post
- Removed "posts per group" for task workflows, i.e. evenly distribute posts to all groups
- Move posts to destination AND remove from source
- Allow submitting each post individually, while will immediately move it to destination and update progress bar

**Things to be broken into separate issues:**
- New class of task workflows requiring students to create posts rather than just reviewing current ones
- Add in voting requirement (not really a requirement, more of an option, i.e. showing/hiding vote buttons)
- Allow distributing posts to whole class, e.g. 1 post per person 
- Clean up workflows modal (I was thinking of maybe just giving this a separate page instead since there's a lot of info)

LunarFang416:
@Ammar-T I'm having trouble while trying to test the branch on my end. The frontend keeps on giving nonsense compiler errors (a lot). Is this happening on your end as well?

Ammar-T:
> @Ammar-T I'm having trouble while trying to test the branch on my end. The frontend keeps on giving nonsense compiler errors (a lot). Is this happening on your end as well?

Can you send a screenshot of the errors?

LunarFang416:
@Ammar-T Its giving a lot of weird errors that don't make sense. something like this:
Screen Shot 2022-08-22 at 7 42 50 AM

Ammar-T:
> @Ammar-T Its giving a lot of weird errors that don't make sense. something like this:
> 
> Screen Shot 2022-08-22 at 7 42 50 AM
> 
> 
> 
> 

Hmm, that error usually happens when that specific module isn't defined inside MaterialModule. 

If that isn't the issue, maybe try doing an npm install and restart the app.

JoelWiebe:
> > @Ammar-T Its giving a lot of weird errors that don't make sense. something like this:
> > Screen Shot 2022-08-22 at 7 42 50 AM
> 
> Hmm, that error usually happens when that specific module isn't defined inside MaterialModule.
> 
> If that isn't the issue, maybe try doing an npm install and restart the app.

"npm install" helped remove these errors for me

Ammar-T:
@JoelWiebe Addressed your feedback from previous review!

> This is a super-powerful feature, so glad we have it :)
> 
> Ideas/Suggestions: **Task Workspace: Submitting Posts**
> 
> 1. Progress bar updates before submit button (expected: update after submit button is pressed -- does the current way work better in some regards?)

Progress updates happen on every action (e.g. if upvoted, it shows the progress updated immediately) instead of waiting for submitting the post. I think this may be clearer in case students are getting a lot of work done but just not submitting posts since they don't want to submit and not be able to go back and change up their actions (e.g. undo a tag). Does that make sense? This would give teachers a more genuine way of tracking progress.

> 2. Submitting a post appears to delete it from the source but also removed it from the list view and does not add it to the destination (expected: remain in list view and add to destination)

Updated now to remain in list view only and copy post to destination as well.

> 3. Submitting the post does not remove the post from the source in real-time (browser refresh required)

I assumed this means removing the post from the canvas? Updated to remove post from canvas if submitted via a task workflow.

> 4. Submitting post by one group member does not remove post for another group member (expected: remove the post from peers' deck of posts in real-time too)

Updated to keep posts deck updated across all task workers.

> **Task Workspace UI** 
> 5. Missing a space in actions required (a) “Tag each post atleast once!” AND (b) "Comment on each post atleast once!"

Fixed missing spaces.

> **Manage Workflows Tab** 
> 6. Can we get a confirmation dialogue when deleting workflows

Added!
 
> 7. Does not list task criteria (e.g., add at least 1 tag to each post) 

Added! 

> 8. Does not have “Task Workflow” label (similar to the “Distribution Workflow” label we already have for bucket workflows) 

Updated with new label!

> 9. Once a task workflow is active, is there any way to make it inactive again (i.e., if the teacher accidentally activated it… to hide it from students again)?

Larger issue, will need to be split up into separate issue.

> **Create Task Workflow Tab** 
> 10. Hide Personal boards from destination list

Updated to remove personal boards from destination!

Ammar-T:
@JoelWiebe Added all your suggestions except #4 which I can either get to this week or we can create it as a separate issue in order to get this PR merged asap.

JoelWiebe (approved):

LunarFang416 (approved):

PR #308, opened 2022-06-30
focus on input fields

### LINKED ISSUE #298: Place Focus in Post Editor
When creating a new post on the canvas, add focus to the "title" field when the editor opens.

When clicking on "comments" on an existing post to expand the comments, place the focus on the comment field.

LunarFang416:
## Details

- When creating new post focus in on title field
- When post comments are clicked, focus is panned to the comment field

Closes #298

### DIFF
--- a/frontend/src/app/components/add-post-modal/add-post.component.html
+++ b/frontend/src/app/components/add-post-modal/add-post.component.html
@@ -4,11 +4,11 @@
     <mat-label>Title</mat-label>
     <input
       name="title"
+      [appAutofocus]="true"
       [formControl]="titleControl"
       [errorStateMatcher]="matcher"
       matInput
       autocomplete="off"
-      cdkFocusInitial
       [(ngModel)]="title"
     />
     <mat-error *ngIf="!title"> Title is <strong>required</strong> </mat-error>
--- a/frontend/src/app/components/add-post-modal/add-post.component.ts
+++ b/frontend/src/app/components/add-post-modal/add-post.component.ts
@@ -1,4 +1,4 @@
-import { Component, Inject, ViewChild, ElementRef } from '@angular/core';
+import { Component, Inject } from '@angular/core';
 import { FormControl, Validators } from '@angular/forms';
 import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
 import { Board } from 'src/app/models/board';
--- a/frontend/src/app/components/add-post-modal/add-post.component.ts
+++ b/frontend/src/app/components/add-post-modal/add-post.component.ts
@@ -1,10 +1,4 @@
-import {
-  Component,
-  AfterViewInit,
-  Inject,
-  ViewChild,
-  ElementRef,
-} from '@angular/core';
+import { Component, Inject, ViewChild, ElementRef } from '@angular/core';
 import { FormControl, Validators } from '@angular/forms';
 import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
 import { Board } from 'src/app/models/board';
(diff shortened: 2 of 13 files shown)


frontend/src/app/components/add-post-modal/add-post.component.ts
```diff
@@ -61,6 +70,10 @@ export class AddPostComponent {
     );
   }
 
+  // ngAfterViewInit(): void {
```
JoelWiebe, line 73: Should we remove this commented out section?

JoelWiebe (approved):
Beautiful! Looks good to me.

frontend/src/app/components/add-post-modal/add-post.component.html
```diff
@@ -4,11 +4,11 @@ <h1 mat-dialog-title>Add Post</h1>
     <mat-label>Title</mat-label>
     <input
       name="title"
+      [appAutofocus]="true"
       [formControl]="titleControl"
       [errorStateMatcher]="matcher"
       matInput
       autocomplete="off"
-      cdkFocusInitial
```
Ammar-T, line 11: So why are we not just using `cdkFocusInitial` for the initial autofocus? What does `[appAutofocus]` do that's better?

LunarFang416: With the cdkFocusInitial it pings the mat-error and makes the title field red when you open the new post modal. I tried to add more conditions like making sure the mat-error triggers once the user has actually typed something, but it still gave the same error, so i went with the directive.
The directive is very simple and straightforward and basically does what the "stock" autofocus html option does (but the problem with that is it only works the first time you open a modal, and not for the times after you have opened the modal once).

frontend/src/app/components/add-post-modal/add-post.component.ts
LunarFang416: Ye, thats my bad i was trying some different ways and forgot to remove this.

frontend/src/app/components/add-post-modal/add-post.component.html
Ammar-T: If the issue persists after the first time you open the modal, then it's probably not the best solution.

Can you maybe try what's suggested here: https://stackoverflow.com/questions/51503940/how-to-avoid-mat-form-field-validation-trigger-when-matinput-lose-focus

LunarFang416: So you want to drop the directive?

Ammar-T: Yeah that would be nice since ideally we want a solution that works when the modals opened multiple times.

LunarFang416: no the directive works perfectly, i mentioned 'autofocus' since its a built in option in html that only works the first time. That lead me to make the `[autoFocus]` directive which works like we want it too.

LunarFang416: j re-read my comment and it sounds like i was talking about the directive only working once xd. i meant the html autofocus.

frontend/src/app/components/add-post-modal/add-post.component.ts
```diff
@@ -32,6 +39,8 @@ export interface AddPostDialog {
   styleUrls: ['./add-post.component.scss'],
 })
 export class AddPostComponent {
+  @ViewChild('titleInput', { static: true })
+  titleInputElement!: ElementRef<HTMLInputElement>;
```
Ammar-T, line 43: Can we also remove this since it was used by the commented function which I don't think we need anymore.

Ammar-T (changes requested):

frontend/src/app/components/add-post-modal/add-post.component.html
Ammar-T: Oh gotcha, yeah I misinterpreted your comment. I'm good to keep this then.

frontend/src/app/components/post-modal/post-modal.component.html
```diff
@@ -176,7 +176,7 @@
       *ngIf="showComments && canStudentComment"
     >
       <mat-label>Add Comment</mat-label>
-      <input matInput [(ngModel)]="newComment" (keyup.enter)="addComment()" />
+      <input matInput [appAutofocus]="true ? commentPress : false" [(ngModel)]="newComment" (keyup.enter)="addComment()" />
```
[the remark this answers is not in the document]
LunarFang416: i fixed this, j need to commit the changes after cleaning up.

PR #309, opened 2022-06-30
tagID for new tags

### LINKED ISSUE #226: Add tag ID for each tag
Create new tags with a unique ID

LunarFang416:
## Details

- TagID created for all new tags added to a board

Closes #226

### DIFF
--- a/backend/src/models/Tag.ts
+++ b/backend/src/models/Tag.ts
@@ -19,6 +19,9 @@ export class TagModel {
   @prop({ required: false })
   public boardID!: string;
 
+  @prop({ required: false })
+  public tagID!: string;
+
   @prop({ required: true })
   public name!: string;
 
--- a/backend/src/models/Tag.ts
+++ b/backend/src/models/Tag.ts
@@ -16,10 +16,10 @@ export class TagSpecialAttributes {
 
 @modelOptions({ schemaOptions: { collection: 'tags' } })
 export class TagModel {
-  @prop({ required: false })
+  @prop({ required: true })
   public boardID!: string;
 
-  @prop({ required: false })
+  @prop({ required: true })
   public tagID!: string;
 
   @prop({ required: true })
--- a/frontend/src/app/models/tag.ts
+++ b/frontend/src/app/models/tag.ts
@@ -7,7 +7,7 @@ export class TagSpecialAttributes {
 
 export class Tag {
   boardID: string;
-  tagID?: string;
+  tagID: string;
   name: string;
   color: string;
   specialAttributes?: TagSpecialAttributes;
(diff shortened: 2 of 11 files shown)


JoelWiebe:
@LunarFang416 Great work!

Can you add the TagID to the trace data as well? We already have postTagNameAdded and postTagNameRemoved but are missing post postTagIDAdded and postTagIDRemoved -- can you also log the ID whenever the name is being logged? Each tag being added or removed by a user adds a row to the trace logs. The variable [dictionary and template of headers is here](https://docs.google.com/spreadsheets/d/1pMqKYt-hc7AQ4kvP2Ey4pQCTNvijqAFN/edit#gid=153344098) (showing the order of the new columns). 
Screen Shot 2022-06-30 at 5 11 51 PM


Logging of trace data can be enabled in the project configurations and downloaded from the project dashboard:
Screen Shot 2022-06-30 at 5 08 25 PM

Screen Shot 2022-06-30 at 5 08 56 PM

Thanks for your work on this. Let me know if there is something that I can further clarify!

backend/src/models/Tag.ts
```diff
@@ -19,6 +19,9 @@ export class TagModel {
   @prop({ required: false })
   public boardID!: string;
 
+  @prop({ required: false })
```
Ammar-T, line 22: This should be required and so should boardID which I'm not sure why it isn't already lol.

frontend/src/app/models/tag.ts
```diff
@@ -7,6 +7,7 @@ export class TagSpecialAttributes {
 
 export class Tag {
   boardID: string;
+  tagID?: string;
```
Ammar-T, line 10: Shouldn't be an optional attribute, the id should be required from now on.

frontend/src/app/utils/FabricUtils.ts
```diff
@@ -69,6 +69,7 @@ export class FabricUtils {
     return DEFAULT_TAGS.map((tag) => {
       const fullTag: Tag = {
         boardID,
+        tagID: tag?.tagID,
```
Ammar-T, line 72: Once you make it required, you can also remove the ? from here.

LunarFang416: Ye will do that, it was giving and error before but now i realize its because i need to updat ethe DEFAULT_TAGS so that they include the tagID as well.

LunarFang416:
@JoelWiebe added tracing for tagID adding and removal. @Ammar-T added tagID fro default tags and made it required.

JoelWiebe (approved):
Fantastic!! Glad you were able to figure out all the myriad info I sent your way :)

PR #311, opened 2022-07-01
Issue 300 - Fix upvote update issue

### LINKED ISSUE #300: Fix Remaining Upvotes
1) When the number of upvotes is increased in the board configurations, the remaining upvotes indicator is successfully updated. However, when the number of upvotes is decreased in the board configurations the number of remaining upvotes is NOT decreased (allowing the number of remaining to get out of synch)->
Screen Shot 2022-06-28 at 2 33 06 PM

2) Could we also make a wording change in the snackbar notification, from "You have already surpassed the upvote limit" to "You have reached the upvote limit"

Screen Shot 2022-06-28 at 2 36 20 PM

zarifmahmud:
## Details

- Fixes bug where teacher updating the upvote limit wouldn't propagate correctly to students without refreshing.
- Change upvote limit message.

Closes #300

### DIFF
--- a/backend/src/api/upvotes.ts
+++ b/backend/src/api/upvotes.ts
@@ -15,7 +15,7 @@ router.post('/', async (req, res) => {
   );
   const board: BoardModel | null = await dalBoard.getById(upvote.boardID);
   if (board && usersUpvotes.length >= board.upvoteLimit) {
-    return res.status(406).end('You have already surpassed the upvote limit!');
+    return res.status(406).end('You have reached the upvote limit!');
   }
 
   const savedUpvote = await dalVote.create(upvote);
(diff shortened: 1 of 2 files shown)


JoelWiebe (commented):
Thanks, looks like the problem is solved!

I seem to be getting some latency on the upvote. Is this just on my end? @zarifmahmud @Ammar-T Does this implementation wait for some kind of callback or is it updated immediately on the canvas?

zarifmahmud:
On the student's end it has to wait for the socket notification to be propagated to it, but other than that I don't think there should be much difference between the teacher's and student's wait times. Is the latency more than other for events, like seeing another student's upvote?

JoelWiebe (approved):
Must have just been some network latency on my end as I was at the cottage. All good -- great work!!

PR #316, opened 2022-07-06
Fix anonymity bug and author name leaks

zarifmahmud:
## Details

- I was getting a bug where if a board is supposed to be anonymous,  refreshing the page causes some posts to be anonymous and others to have the correct name, like in this picture:

image

To recreate:

1. Create a new board using teacher, and have a student account join it.
2. Make some posts with the student account.
3. Teacher account unchecks the permission boxes that allows teachers and students to see post name, and updates the configuration. (this can also be done now, or in the very beginning)
4. Then if you access the board from any account (teacher, student who made the posts, or another student), and keep refreshing, you'll reach a state where one or more of the posts aren't anonymized.

- Also, refreshing the page could lead to an author's name being shown for a split second before it switched to anonymous.
- This fix is meant to fix the bug and the author name leak.

### DIFF
--- a/frontend/src/app/components/canvas/canvas.component.ts
+++ b/frontend/src/app/components/canvas/canvas.component.ts
@@ -336,6 +336,7 @@ export class CanvasComponent implements OnInit, OnDestroy {
   }
 
   configureBoard() {
+    console.log('configuring board');
     const map = this.activatedRoute.snapshot.paramMap;
 
     if (map.has('boardID') && map.has('projectID')) {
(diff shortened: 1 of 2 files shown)


zarifmahmud:
The fix is a bit weird, since I'm adding an optional "authors" param to the PostOptions, but the thinking behind it is when a board is first loaded it will now by default show author names as "Anonymous", and then later on in `configureBoard()` when `this.setAuthorVisibilityAll()` is called, the author names can be loaded.

JoelWiebe (commented):
Successful tests:
- Adding/removing anonymity privileges makes immediate change on both teacher and student canvases 
- Refreshing the browser maintains anonymity 

Additional items/questions
- posts in lists and buckets do not appear to get anonymized
- do you think it makes sense that users should still be able to see display name for their own posts and anonymize only others' posts (this way if they wanted to locate their posts they could)

PR #317, opened 2022-07-06
Clear Board Functionality 

### LINKED ISSUE #159: "Clear Board" button
Allow users to clear the canvas which means removing all posts currently on the canvas. 

However, this should keep those posts visible in the list view or bucket view (if it was assigned to buckets), so it's not completely gone. 

Add a permission for who can see this button (i.e. only for teachers, or both)

LunarFang416:
## Details

- Clear Board functionality implemented. Button located in board configuration:
Screen Shot 2022-07-06 at 11 00 37 AM

- Created Trace for clearing board which is triggered by new eventType: `BOARD_CLEAR`

Closes #159


frontend/src/app/components/add-board-modal/add-board-modal.component.ts
```diff
@@ -56,6 +56,7 @@ export class AddBoardModalComponent implements OnInit {
       allowStudentEditAddDeletePost: true,
       allowStudentCommenting: true,
       allowStudentTagging: true,
+      // allowStudentClearBoard: true,
```
JoelWiebe, line 59: Can we remove this comment?

JoelWiebe (commented):
So cool!! 


- Clear Board button successfully removes items from the board but they remain in buckets and list view
- Board_Clear trace log looks good

Just one small thing. Could we specify in the Confirmation dialogue that the posts are not being deleted? E.g., 

"Are you sure you want to clear posts from this board? NOTE: Posts will be cleared from the board but remain in the the list view and any assigned buckets."

P.S. I noticed that (1) I could not download posts from the list view (clicking the download button has no effect in the list view, but it works in the buckets) and (2) there was no create new post button in the list view. Had those fixes been completed or merged into this branch?

This is another really neat feature. Thanks for your work!

LunarFang416:
@JoelWiebe for downloading the posts from the list view, i fixed that in the Delete bucket PR. For creating posts in the list view, i think thats feature hasn't been implemented yet. Ill create a new issue for that and work on that in a separate PR.

JoelWiebe (approved):

PR #318, opened 2022-07-06
Open task snack bar functionality 

### LINKED ISSUE #238: Open task snackbar when board is opened
Add a board configuration setting that would open the task snackbar each time someone opens the board. Teachers could turn this feature on or off in the board configuration settings. By default this should be turned off.

LunarFang416:
## Details
- Opening task snack bar when board is opened permission added
- Permission is false by default
Screen Shot 2022-07-06 at 4 23 50 PM

Closes #238

### DIFF
--- a/backend/src/models/Board.ts
+++ b/backend/src/models/Board.ts
@@ -43,6 +43,9 @@ export class PermissionsModel {
   public showBucketStudent!: boolean;
 
   @prop({ required: true })
+  public showSnackBarStudent!: boolean;
+
+  @prop({ required: true })
   public allowTracing!: boolean;
 }
 
--- a/frontend/src/app/components/add-board-modal/add-board-modal.component.html
+++ b/frontend/src/app/components/add-board-modal/add-board-modal.component.html
@@ -133,6 +133,11 @@
           >Show buckets to students</mat-checkbox
         >
       </p>
+      <p>
+        <mat-checkbox [(ngModel)]="permissions.showSnackBarStudent"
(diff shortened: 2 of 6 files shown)


backend/src/models/Trace.ts
```diff
@@ -27,6 +27,7 @@ export class TraceModel {
   eventType!: string;
 
   @prop({ required: true })
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
JoelWiebe, line 30: Remove?

JoelWiebe:
Suggestions:
- Can we change the setting label from "Open task snackbar when board accessed" to "Display task description when board is opened"
- In the snackbar, can we also remove the "!" from the buttons for Close and View Full Task
- Let's move "Display task description when board is opened" and "Allow tracing of board activity" from the Permissions tab (intended for controlling student permissions) to the first "Board" tab (intended for board settings)

Perhaps we can place the two settings options either just above initial Zoom or just below Max upvotes -- whichever seems to look more balanced
Screen Shot 2022-07-09 at 3 41 08 PM

backend/src/models/Trace.ts
LunarFang416: thats there to disable the linter warning.

LunarFang416:
@JoelWiebe how does this look?

Screen Shot 2022-07-11 at 11 26 53 AM

Screen Shot 2022-07-11 at 11 26 58 AM

JoelWiebe (approved):

PR #328, opened 2022-07-10
Grouping UI Draft

DCH25:
## Details

- Most of the functionality is there (creating, updating, deleting), still has some bugs/needs some code cleanup
- Need to add the assignment tool (for switching members between multiple groups), for that I'm changing the group-member-list component to support a variable amount of groups
- Going to switch from the checkboxes in the group creation menu to the drag and drop tool (seen in group editing) in a later commit


JoelWiebe:
Really cool drag and drop, great work! I also like the way you did the multi-level menu when editing group members.

A few suggestions/ideas:

1. For simplicity, I think we could just have 2 tabs and move "create" group to the first tab (see Figure 1)
2. Our groupings should allow students to exist in multiple groups so unassigned for a group should be any students not in that group
3. Let's change that name of that Assign tab to "Multi-Group Assign)
4. For the Multi-Group Assign tab, can we use essentially the structure that Hiroki used where multiple groups can be selected; one column is displayed for each group; in this case, the unassigned column shows the students who are not in any of the selected groups (but could theoretically be in other groups) [Hiroki's code for manual assignment](https://score.oise.utoronto.ca/teacher/manage/unit/82/manage-groups) -- please do reach out if you have any questions about functionality
5. If random assignment will take a bit of time, we can always do that in another PR
6. Generally speaking, was it possible to port over what Hiroki had created or was it easier to build a lot of these components from scratch?

**Figure 1.** Move Create Group
Screen Shot 2022-07-11 at 10 08 29 PM

**Figure 2.** Use Hiroki-like Multi-assign
Screen Shot 2022-07-11 at 9 57 53 PM

Let me know if you ever want to have a quick meeting to discuss anything.

Keep up the good work!

DCH25:
Awesome thanks for the feedback! 

4. For the Multi-Group Assign tool, do we want it to scale the same way as score (i.e. modal grows to fit the number of groups, see in image below) or do we want something like the [multiple selection](https://material.angular.io/components/select/overview) from materialUI? 
5. For shuffling I can just take the logic from Hiroki's code, so it shouldn't take too long
6. So far it's been easier to write from scratch and use the score client for reference

image

PR #331, opened 2022-07-12
Clear Vote Functionality

### LINKED ISSUE #324: Add "Clear Votes" button in Board Config
In the board configurations, add a "Clear Votes" button that removes all dot-votes from that board. This should also add a "VOTES_CLEAR" event in the trace logs (similar to BOARD_CLEAR)

LunarFang416:
## Details

- Clear Vote functionality implemented in the board configuration
- Clear Vote logs the `VOTE_CLEAR` trace
- Sockets implemented to update vote clearing across all active boards in real time

Closes #324


frontend/src/app/components/canvas/canvas.component.ts
```diff
@@ -118,6 +119,7 @@ export class CanvasComponent implements OnInit, OnDestroy {
       [SocketEvent.BOARD_TAGS_UPDATE, this.handleBoardTagsUpdateEvent],
       [SocketEvent.BOARD_TASK_UPDATE, this.handleBoardTaskUpdateEvent],
       [SocketEvent.BOARD_UPVOTE_UPDATE, this.handleBoardUpvoteUpdateEvent],
+      [SocketEvent.VOTES_CLEAR, this.handleVotesClearEvent],
```
Ammar-T, line 122: I think you also need to add a listener for this inside `list-modal.component.ts` and `buckets-modal.component.ts`.

JoelWiebe:
Awesome job!

Tested:
- Clear student and teacher votes in real-time

1. Opened a board with both student and teacher accounts
2. Voted with both student and teacher
3. Clicked the Clear Votes button
-> Both teacher and student votes cleared and the total number of remaining votes for both were reset to maximum amount
-> Logged Clear_Votes event

Notes:
- I actually couldn't find the Clear Votes button initially (I was refreshing my browser, pulling and resetting with git) then finally realized I just needed to scroll down. Is there some way we can implement a "Scroll to Bottom" button or somehow indicate that there are more UI elements lower in the page? If this is a big task, let's address the fix in another issue.

This is what I was seeing:
Screen Shot 2022-07-15 at 10 11 52 AM

LunarFang416:
@JoelWiebe Its happened to me as well, and I have to scroll all the way down to see the buttons. We could add an indicator to show that there is still something below, or we could try to have buttons side by side which would help in showing more buttons at once. What do you think will be a better fit?

JoelWiebe:
@LunarFang416 Oh my goodness, sorry for my delay. 

As a quick fix we could stack the buttons into 2 columns, but on mobile devices they will probably need to be a single column with some indicator that something more is below. 

Would it work to just make the scroll bar always visible in the board configurations? E.g., [https://www.w3schools.com/howto/howto_css_force_scrollbars.asp](https://www.w3schools.com/howto/howto_css_force_scrollbars.asp)

LunarFang416:
> @LunarFang416 Oh my goodness, sorry for my delay.
> 
> As a quick fix we could stack the buttons into 2 columns, but on mobile devices they will probably need to be a single column with some indicator that something more is below.
> 
> Would it work to just make the scroll bar always visible in the board configurations? E.g., https://www.w3schools.com/howto/howto_css_force_scrollbars.asp

I'll format the buttons into 2 columns. I think this since the blank whitespace takes up a lot of space. 

For the scrollbar, i can see it by default. I'm not sure if thats the case for everyone, or if you can't see it only; please let me know.

Ammar-T (approved):

JoelWiebe (approved):
Looks great!

Side notes:
- I just noticed that on my browser the config buttons stay as 2 columns rather than collapse back to one, but this is okay by me
- Also on my browser, scroll bars only actively show up when I'm scrolling -- which is fine in most cases, but perhaps there is a way to make scroll bars remain visible in some cases -- just something we can keep in mind for other cases (e.g., source/destination dropdown lists for bucket workflows)

This should be good to merge though!

PR #335, opened 2022-07-15
Tag/Upvote Workflows

### LINKED ISSUE #326: Add Tag- and Vote-based Bucket Workflows
In the authoring of bucket workflows, add two new distribution options:
Screen Shot 2022-07-10 at 12 43 26 AM
Screen Shot 2022-07-10 at 12 43 52 AM

Below "random" add options for "by tag" and "by vote"

- If "by tag" is selected, display a drop-down list of available tags for the board
- If "by vote" is selected, display a the label "has at least" followed by a numerical input with a min value of 1 and max value equal to the max value of upvotes set in the board config (i.e., a slider could be used here)

When executed in Manage Workflows, all posts matching the criteria from the source bucket/board will be added to each of the selected buckets/boards
Screen Shot 2022-07-10 at 12 51 40 AM

LunarFang416:
## Details

- Tag and upvote workflows implemented:
Screen Shot 2022-07-14 at 8 04 47 PM

- For tag you will select one tag from all tags that are available in the current board and it will transfer all posts that contain that tag from source to the destination
- For upvotes, you can configure the minimum amount of upvotes per votes which will dictate which posts will  be transferred from the source to the destination boards/buckets. This can be done through a slider. 
- I slightly changed the Manage workflows modal to reflect the new distribution types:

Screen Shot 2022-07-14 at 8 05 14 PM

Closes #326


JoelWiebe:
That's flippin' amazing!! I just quickly pulled it to check it out. Will test more shortly!!!

backend/src/agents/workflow.agent.ts
```diff
@@ -31,34 +33,103 @@ export const runDistributionWorkflow = async (
...
-  let sourcePosts;
-
-  if (source.type == ContainerType.BOARD) {
-    sourcePosts = await dalPost.getByBoard(source.id);
-    sourcePosts = sourcePosts.map((p) => p.postID);
-  } else {
-    const bucket: BucketModel | null = await dalBucket.getById(source.id);
-    sourcePosts = bucket ? bucket.posts : [];
-  }
+  if (
```
Ammar-T, line 36: Can we split each distribution type's code into different smaller functions? i.e. have a function for `randomDistribution`, `tagDistribution`, etc..

JoelWiebe:
@LunarFang416 
Just noticed a couple things:

- Can we add the board as one of the options for both source and destination (or even just the source)?
- Can you clear the input fields on the Create Workflow tab after the workflow is successfully created?
- For some reason tag distribution does not send posts to the first destination selected (only the 2nd, 3rd, etc.) -- everything seems to work fine for upvote distributions
- If you have time, can you add a toggle option in workflow creation and management for “Remove from source” (Yes/No)

LunarFang416:
> @LunarFang416 Just noticed a couple things:
> 
> * Can we add the board as one of the options for both source and destination (or even just the source)?
> * Can you clear the input fields on the Create Workflow tab after the workflow is successfully created?
> * For some reason tag distribution does not send posts to the first destination selected (only the 2nd, 3rd, etc.) -- everything seems to work fine for upvote distributions
> * If you have time, can you add a toggle option in workflow creation and management for “Remove from source” (Yes/No)

1. The board name is included in the source options already, and other boards in the same project are included in the. destination list
2. Done.
3. I tested this a couple times and the workflow was sending all required posts to all buckets. I also double checked the code for any discrepancies, and there was not an issue there
4. Done.

Please let me know if its all working as you expect it.

JoelWiebe:
@LunarFang416 
RE:
1. Sorry, I must have somehow missed the board being in the source list (I think my board name was similar to my bucket names)
3. Somehow I'm still getting an issue, for example if I move all posts from one bucket to one other bucket then no buckets are moved. Can you try the workflow I created here:

- There are 6 posts in Bucket 1
- All post are tagged with Idea
- The workflow should move all 6 posts tagged with Idea from Bucket 1 to Bucket 2 and remove them from Bucket 1
- When I run the workflow, all 6 posts are removed from Bucket 1 but not added to Bucket 2

Can you try this workflow to see if you're getting the problem on your end as well? Project invite code: 6QZ7U

LunarFang416:
@JoelWiebe Just tested it out and I'm having the same error. Just made some small tweaks in the code, and tested again and it works fine. And I hope thats the case on your end as well :).

JoelWiebe:
@LunarFang416 Awesome, I like that idea of leaving the posts in the listview when they are removed from the board -- that way we can always find all non-deleted notes in the listview if we want to move them somewhere else or download them back to the canvas. 

I found one last item, then we should be good to go from a UX POV :) When we remove posts from the board, I notice that they don't disappear right away from the board. I need to refresh my browser for them to disappear. Could we have moved posts removed from the board right away?

Great work on these features!!

LunarFang416:
@JoelWiebe, just made a few changes and the backend now emits a socket event which removes posts from board/bucket (if remove from post is selected) in real time.

JoelWiebe (approved):
B-E-A-UTIFUL, thanks!

backend/src/repository/dalPost.ts
```diff
@@ -23,6 +23,15 @@ export const getByBoard = async (boardID: string) => {
   }
 };
 
+export const getBoardTypePosts = async (boardID: string) => {
```
Ammar-T, line 26: Instead of creating a new function for querying boards by type, can we update the current `getByBoard` to take an optional parameter `type?: PostType` and if it's specified by the caller, we just include it in the query?

Ammar-T: After this fix, feel free to merge!

Ammar-T (approved):

backend/src/agents/workflow.agent.ts
```diff
@@ -59,9 +63,136 @@ export const runDistributionWorkflow = async (
...
+      await dalBucket.addPost(destination.id, filteredPosts);
+    }
+  }
+
+  if (removeFromSource) {
+    return removePostFromSource(source, filteredPosts);
+  }
+};
+
+export const runDistributionWorkflow = async (
```
Ammar-T, line 157: Actually, one more thing - this should be renamed, i.e. `runWorkflow`

PR #345, opened 2022-08-16
README for repo

### LINKED ISSUE #323: Add README
Any thoughts on adding a README for the repo since there was one before the websockets implementation. We should also remove the README from the /frontend folder since its outdated (or maybe tweak that a little bit and use it as the README for the repo).

LunarFang416:
## Details

- Tweaked the existing readme in `/frontend` and move it too root

Closes #323

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,36 +4,59 @@ This Common Knowledge (CK) Board is the latest rebuild of the ENCORE Lab's colla
 
 ## Setup
 
-#### 1. Install Node.js and Angular
+### 1. Install Node.js and Angular
 
 - Node: https://nodejs.org/
   - Use the LTS versions. v16 works, v17 has some compatabily issues
 - Angular: `npm install -g @angular/cli `
 
-#### 2. Clone the repository
+### 2. Clone the repository
 
 ```shell
 $ git clone https://github.com/encorelab/ck-board.git
 ```
 
-#### 3. Install required dependencies
+### 3. Install required dependencies
 
-If you are building locally or on an instance install depedancies with
-
-```shell
-$ npm ci
-```
-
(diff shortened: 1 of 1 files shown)


JoelWiebe (commented):
@LunarFang416 Thanks for moving and updating the README!

Did we want keep our .env file values private or do we not care if these are available to the public? I suppose since this is not for our production server, it's not a big deal... but just curious what best practice would be :)

Ammar-T:
@JoelWiebe Good point Joel, I think it'd be better to keep those keys hidden even if it's not prod.

LunarFang416:
Yeah that makes sense. I wanted to exclude them first, but was unsure so included them so that you can also take a look as well.
Will change it now.

README.md
```diff
@@ -36,13 +36,13 @@ $ npm install
 
 Create a `.env` file inside the `/backend` directory (`touch .env` in your terminal or manually create file using your IDE)
 
-Add the following content into the `.env` file:
+Add the following content into the `.env` file (You will have the required credentials if you area authorized to continue development on this project):
```
JoelWiebe, line 39: Is there a way we could write this so that anyone knew the accepted types of values so they knew how they could swap in their own? E.g., 

Add the following content into the `.env` file (entering your own corresponding Mongoose and JWT values)
DB_USER=[mongoose db user]
DB_PASSWORD=[mongoose db password]
DB_NAME=[mongoose db name]
JWT_SECRET=[JSON web token secret]

(maybe re-word a bit if those aren't the best descriptors)

JoelWiebe (approved):
B-E-A-UTIFUL

Ammar-T (approved):

PR #388, opened 2022-11-02
Issue 353 - Bucket Post Create Log

### LINKED ISSUE #353: Add "Create" log on post creation from buckets
The creation of a post from a bucket view should trigger a trace log of both "creating" and "adding" a post to this bucket. 

On creation of a post from a bucket view, a BUCKET_ADD_POST event is logged in trace logs. This is good, but we are missing the additional POST_CREATE event that includes the postModifiedTitle and postModifiedMessage values

DCH25:
## Details

- Small change: add a POST_CREATE trace along with the BUCKET_ADD_POST trace when adding posts from the bucket modal.

Closes #353


JoelWiebe (approved):
Thanks! You're free to merge this :)

Issue #400, opened 2022-11-10
Trace log bugs


JoelWiebe:
I'm noticing a couple issues with trace logging on the production server (should be same as develop branch)
 - Downloading trace logs appears to be limited to the most recent 50 traces
 - Incorrect clientTimestamp -> as can be seen in these trace logs on the production server (teacher invite code MVS3B) many events have the same clientTimestamp but different serverTimestamps
 - There was one observed CREATE_POST event that was generated 4 times for a single new post being created (but haven't been able to replicate this)

PR #409, opened 2022-11-15
Issue 400 - Change trace log limit to 1000

DCH25:
## Details

- Trace log CSV files contain the 1000 most recent logs instead of 50

From #400


DCH25:
Client timestamp updates whenever an event is emitted instead of only when the board is initialized or its permissions are updated.

JoelWiebe (approved):
Perfect! Feel free to merge immediately.

Issue #390, opened 2022-11-03
Add Group Todo List


JoelWiebe:
1. Add a "Group Todo List" tab in the Todo List tool
<img width="496" alt="Screen Shot 2022-11-03 at 5 59 13 PM" src="https://user-images.githubusercontent.com/6416247/199842449-c19c54e0-1bca-4d70-9bd0-183476550231.png">

 - This new tab will display all of the todo list items that contain a group
 - Add an additional column for "Group" within the Group Todo List tab
 - The Group todo list tab should show all todo list items assigned to groups that the current user is currently a member of (i.e., including TODO list items created by other students)

2. Rename the Todo List tab to Personal Todo List
 - This tab will display todo list items for the current user that have NO group assigned
<img width="529" alt="Screen Shot 2022-11-03 at 7 05 39 PM" src="https://user-images.githubusercontent.com/6416247/199851665-6e7623a6-75da-40ab-9a00-50c1968222c5.png">

3. When creating a new todo item from either the Personal or Group Todo List tab, the following pop-up will be shown that allows for creation of items with or without a group
 - group selection should include only groups that the current user is currently a member of
<img width="497" alt="Screen Shot 2022-11-03 at 5 53 23 PM" src="https://user-images.githubusercontent.com/6416247/199841649-2c1c4a3c-0e91-4011-bb9e-943e790fdd61.png">

Issue #381, opened 2022-10-20
Create TODO list types


JoelWiebe:
To add todo list types to the personal todo tool, we will hard-code three types of goals:
 - Cognitive strategies
 - Social-emotional learning supports
 - Behavioural engagement practices
 - Class-related work

1. Add an additional column to the Todo List tab for displaying the item Type
<img width="529" alt="Screen Shot 2022-11-03 at 7 05 39 PM" src="https://user-images.githubusercontent.com/6416247/199851665-6e7623a6-75da-40ab-9a00-50c1968222c5.png">

2. When adding a new todo item, add a "Select Type(s)" multi-select field for choosing one (or more) of the following options:
 - Cognitive strategies
 - Social-emotional learning supports
 - Behavioural engagement practices
 - Class-related work
<img width="497" alt="Screen Shot 2022-11-03 at 5 53 23 PM" src="https://user-images.githubusercontent.com/6416247/199841649-2c1c4a3c-0e91-4011-bb9e-943e790fdd61.png">

3. Hovering over the info icon next to the type selection field, a tooltip will be displayed:
 - Cognitive engagement strategies (i.e., learning strategies for memorization, recall, application of knowledge)
 - Social-emotional learning supports (i.e., supports for developing your math identity, positive motivation, emotional awareness, stress management, relationships, or critical/creative thinking)
 - Behavioural engagement practices (practices for improving learning engagement, i.e., note-taking, asking questions, generating/sharing ideas, or discussing ideas with peers) 

This issue is a pre-requisite for the [Group Todo List issue](https://github.com/encorelab/ck-board/issues/390)

JoelWiebe:
Hey @LunarFang416, just FYI, I updated the type labels and tooltip info.

PR #415, opened 2022-12-01
Issue 414 ck monitor todo

### LINKED ISSUE #414: Add TODOs monitoring to CK Monitor
In the CK Monitor for the teacher, allow for quick monitoring of students' TODOs

1. Create a "View by TODOs" button on the left panel, when clicked the TODO monitor is displayed
Screen Shot 2022-11-24 at 9 47 46 AM

2. Above the TODOs monitor table, include a date filter that filter's the table by deadline day by the provided range (a calendar date picker can be used for entering the date, preferably year/month/day)

3. The table will tentatively have 7 columns
 - Student/Group: display the name for a personal TODO or the group name instead for a group TODO
 - Goal: name of the goal
 - Type: name fo the goal type
 - Status: Bolded red if the deadline has been missed, bolded green if it has been completed, or bolded dark yellow/orange if it is pending (it is not done, but the deadline has not yet passed) -- I suppose this could be calculated on the client side, given the deadline date and time
 - Deadline (preferably year/month/day)
 - Rubric Score (blank for now)
 - Completion Notes (blank for now)

4. Table should be sortable by column but default to descending by deadline date

DCH25:
## Details

- Added project todo monitor in the ck-monitor page
- Can be sorted by column or filtered by deadline date range
- Only supports individual todos without a type for now (see #381 & #390)

Closes #414


JoelWiebe (commented):
Brilliant work!

Tested:
 - Add single person TODOs
 - Columns are sortable
 - Missed status if deadline passed
 - Pending status for upcoming, incomplete items
 - Complete status for all completed items

Suggestions:
 - Could we do something like the image below where the "Set as Default" button (1) applies that value and (2) saves the date range for all users (this will be the default range when any user opens this tool)? This way the teacher does not need to continually reset the range if often only wanting to view items for a particular class unit. The "Apply" button can continue to be used as it is now to temporarily override the date range (if the teacher would want to quickly see a different range but not change the saved default value).
Screen Shot 2022-12-06 at 1 30 42 PM
 - Can we add even a few pixels of cell padding (currently padding is set to zero)
Screen Shot 2022-12-06 at 3 08 59 PM
 - Can we show either TODOs or Tasks (not both)? For instance, when Monitor Task is clicked can we hide TODOs; and when View by TODOs is clicked can we hide all task monitors?

JoelWiebe:
@LunarFang416 @Ammar-T Would either of you have a chance to review this? We may have a chance to run this on Wednesday/Thursday this week. If so, I will create a new issue containing my suggestions and we will merge the current features. No pressure if your week is already full!

backend/src/models/Post.ts
```diff
@@ -1,12 +1,28 @@
...
 
+export class MultipleChoiceOptions {
+  @prop({ required: true })
+  public optionTitle!: string;
+
+  @prop({ required: true })
+  public correct!: boolean;
+
+  @prop({ required: false })
+  public formuala?: boolean;
```
Ammar-T, line 23: Just noticed this typo, we should have this fixed in another PR since I think this code is from an older task.

frontend/src/app/components/ck-monitor/ck-monitor.component.ts
```diff
@@ -168,6 +214,46 @@ export class CkMonitorComponent implements OnInit, OnDestroy {
...
   }
 
+  async updateTodoItemDataSource(): Promise<void> {
+    const data: TodoItemDisplay[] = [];
+
+    for (const item of this.todoItems) {
+      const date = new Date(`${item.deadline.date} ${item.deadline.time}`);
+      const formattedDate = date.toLocaleDateString('en-CA');
+      const currentDate = new Date();
+      const name = await this.userService.getOneById(item.userID);
```
Ammar-T, line 224: This looks to be pretty expensive, we'd be making a call for each Todo. I would suggest to be able to pass in an optional parameter through the `this.todoItemService.getByProject(projectID)` request which would return a more "expanded" version of each todo where the username is already loaded. We've done this with a couple things already, here's an example: https://github.com/encorelab/ck-board/blob/develop/backend/src/api/workflows.ts#L223

JoelWiebe: Could we move this to another issue to rush the merge for this run or would you advise against going ahead with the merge (currently we have 17 students each with just a couple - probably 1 or 3 - TODO items)?

Ammar-T: Based on those numbers, I think its fine to merge now!

PR #433, opened 2022-12-17
ISSUE-432: Do not create trace event when tracing disabled

### LINKED ISSUE #432: BUG: Creating trace event when tracing disabled
Currently, if tracing is disabled, the backend still attempts to create a trace event when comments are removed.

Ammar-T:
## Details

- Checks if tracing is enabled before attempting to create a trace event for comment deletions.

Closes #432

### DIFF
--- a/backend/src/socket/events/post.events.ts
+++ b/backend/src/socket/events/post.events.ts
@@ -178,7 +178,8 @@ class PostCommentRemove {
   ): Promise<object> {
     const comment = input.eventData;
     const commentAmount = await dalComment.getAmountByPost(comment.postID);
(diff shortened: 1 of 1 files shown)


JoelWiebe (approved):
Good catch! This seems to work for me now. Good to merge.

Let's try this out on the production server when we can. Strangely, for the board on our production server that was having this similar bug, trace logs were enabled. Do you think there might be another issue in that case?

Ammar-T:
Hmm, yeah that sounds to be a different issue most likely. We can continue to try to investigate if there's anything else that breaks the submit/commenting.

PR #461, opened 2022-12-24
ISSUE-413: Change post colour based on author's role

### LINKED ISSUE #413: Change teacher post colour for personal boards
Change the background colour (e.g., blue or some other colour that roughly fits into the existing colour palette but ensures the post text is still readable) for teachers' posts (so they stand out) across all canvases, list views, bucket views, but NOT in CK Workspace

Ammar-T:
## Details

- Assign yellow background to student posts and blue background to teacher posts.
- Implemented for all posts across canvas, list view, bucket view and workspace view.

Closes #413

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,12 @@ $ cd frontend    # Go into frontend folder
 $ ng serve --port 4201      # Run Angular app; Will start application on http://localhost:4201/
 ```
 
+## Running database migrations
+
+Switch to the `/backend` directory.
+
+
+
 ## Running unit tests
 
 Run `ng test` to execute the unit tests via [Karma](https://karma-runner.github.io).
--- a/backend/migrate-mongo-config.ts
+++ b/backend/migrate-mongo-config.ts
@@ -1,0 +1,19 @@
+import dotenv from 'dotenv';
+dotenv.config();
+
+console.log(process.env.DB_URI);
+
+export = {
+  mongodb: {
+    url: process.env.DB_URI,
+    databaseName: process.env.DB_NAME,
+    options: {
+      useNewUrlParser: true, // removes a deprecation warning when connecting
+    },
+  },
+  migrationsDir: 'migrations',
+  changelogCollectionName: 'changelog',
+  migrationFileExtension: '.ts',
+  useFileHash: false,
+  moduleSystem: 'esm',
+};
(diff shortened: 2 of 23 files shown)


JoelWiebe:
Thank for this!

Tested:

 - Teachers' posts in personal board, community board, list view, buckets, and workspace should all be coloured blue
 - Students' posts should remain, original yellow in all contexts

Suggestions:

 - Can we have old teachers' posts displayed in new blue (to support previous projects)? I notice that while new teacher posts are displayed in blue old posts in the list view and workspace have no background colour
 - Can we update the original yellow to rgb(255, 230, 120) and new blue to rgb(40, 200, 255)?
Screen Shot 2022-12-28 at 1 51 53 PM

JoelWiebe:
@LunarFang416 Would you mind reviewing this one?

JoelWiebe:
@Ammar-T Do you have a quick moment to wrap this one up?

Ammar-T:
@LunarFang416 Feel free to review when possible.

@JoelWiebe As discussed, changed colour constants and will be doing migration in another PR.

JoelWiebe:
Thanks @Ammar-T, after this has been reviewed, perhaps we can hold off on merging this into develop until the migration PR is also ready to go so that we don't have blank coloured posts.

LunarFang416 (approved):
LGTM

JoelWiebe:
@Ammar-T What would be involved in the migration? Could we just manually update the values in the DB for the one active run we have going now that uses personal boards? That way we could get this merged in!

Ammar-T:
@JoelWiebe I added the migration script in this PR which sets the appropriate post colors for all boards. But, I haven't actually ran the script for all boards in production/staging yet. I did run it for boards in the project you specified (the link you sent on slack). So, you can check the boards in that project and see that teacher posts are now blue. NOTE: If you move around those posts, the color will change back to yellow since this feature isn't merged into production yet. The only reason the posts are blue is because the script was ran (i.e. the script iterated over all those posts and simply changed their color for now).

The reason I haven't ran the script for all projects is because I think we should have some formal process of running migrations. For example, everytime we merge a pull request and there was a new migration added, the migration would run immediately after and keep things stable. The same would be for production. Everytime we create and deploy a production build to Azure, we should run the migrations on the production database to keep things stable there as well.

The thing I haven't figured out yet is how to run these migrations during the build stage/after PR merges. It would have to be done via GitHub actions I think, but it needs to be looked into more, so I think it's worth separating into a new issue.

JoelWiebe (approved):
Looks good.

Perhaps there is some limitation preventing us from doing so, but something to consider for the the future is that we should try to hard set a class rather than hard setting the color for each post. This way we could very easily change the colour of all teacher posts without updating all the colour values stored for each previous posts.

PR #383, opened 2022-10-20
Changes made to Create Workflow Modal

LunarFang416:
## Details

View is now like this:
Screen Shot 2022-10-20 at 5 19 53 PM

Closes #375

PR #486, opened 2023-01-28
feat(canvas): Prevent event propagation of board scroll and pan

breity:
## Details
When user scrolls or pans a board with the mousewheel or trackpad, event propagation is stopped. This should prevent the parent of the iframe that an embedded board is loaded in from also scrolling.

Closes #383.


JoelWiebe (approved):
Nice, clean fix!

Good to merge.

PR #489, opened 2023-01-31
Re-Activate/Cancel Tasks

### LINKED ISSUE #467: Cancel or Reopen Active Tasks
1. In the CK Monitor, create a Cancel option for unfinished tasks
Screen Shot 2022-12-31 at 3 52 38 PM

- [x] For active tasks, create an "Action" column on the far right of each row for tasks
- [x] For active tasks, add a "Cancel" icon to each row
- [x] For the Cancel icon, add the tooltip "Cancel task"
- [x] When the button is pressed, move the task from the Active Tasks list to the Completed Tasks list

2. In the CK Monitor, create a Play icon to send tasks back to Active tasks
Screen Shot 2022-12-31 at 3 53 27 PM

- [x] For completed tasks, create an "Action" column on the far right of each row for tasks
- [x] For completed tasks, add a "Play Circle" icon to each row
- [x] For the Play Circle icon, add the tooltip "Re-activate task"
- [x] When the button is pressed, move the task from the Completed Tasks list to the Active Tasks list

LunarFang416:
## Details

- From the CK-Monitor, teachers can now Cancel Active tasks, which will move the specified group into the "Completed Task" area
- Likewise, Group Tasks in the "Completed task" area can be activated and will move to the "Active Task" area
- No additional changes for Inactive tasks

Closes #467

### DIFF
--- a/backend/src/api/workflows.ts
+++ b/backend/src/api/workflows.ts
@@ -327,25 +327,14 @@ router.post('/task/groupTask/:groupTaskID/submit', async (req, res) => {
 });
 
 /**
- * Mark group task as complete.
+ * Update Group Task status to Complete/Active
  */
-router.post('/task/groupTask/:groupTaskID/complete', async (req, res) => {
+router.post('/task/groupTask/:groupTaskID/status', async (req, res) => {
   const { groupTaskID } = req.params;
+  const { status } = req.body;
 
   const updatedGroupTask = await dalGroupTask.update(groupTaskID, {
-    status: GroupTaskStatus.COMPLETE,
-  });
-  res.status(200).json(updatedGroupTask);
-});
-
-/**
- * Mark group task as active.
- */
-router.post('/task/groupTask/:groupTaskID/active', async (req, res) => {
-  const { groupTaskID } = req.params;
-
-  const updatedGroupTask = await dalGroupTask.update(groupTaskID, {
-    status: GroupTaskStatus.ACTIVE,
+    status: status,
   });
   res.status(200).json(updatedGroupTask);
 });
--- a/backend/src/api/workflows.ts
+++ b/backend/src/api/workflows.ts
@@ -338,4 +338,16 @@ router.post('/task/groupTask/:groupTaskID/complete', async (req, res) => {
   res.status(200).json(updatedGroupTask);
 });
 
+/**
+ * Mark group task as active.
+ */
+router.post('/task/groupTask/:groupTaskID/active', async (req, res) => {
+  const { groupTaskID } = req.params;
+
+  const updatedGroupTask = await dalGroupTask.update(groupTaskID, {
+    status: GroupTaskStatus.ACTIVE,
+  });
+  res.status(200).json(updatedGroupTask);
+});
+
 export default router;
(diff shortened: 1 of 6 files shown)


JoelWiebe (changes requested):
This feature is very helpful, thanks for all your work on this! Most of this works great, I just noticed a couple small things. 

Suggestions:
1. Table padding

- [ ] Can we add cell padding in the table? For instance, I notice that the "Progress" and "Action" headings can become pressed closely to one another (see image below)

Screen Shot 2023-01-31 at 2 02 15 PM

2. Console logs

- [ ] Can we remove the table logs in the console, probably from testing (see image below)

Screen Shot 2023-01-31 at 2 10 50 PM

3. There is a bug when a task is assigned to multiple groups and some are still Pending while others are Active, e.g.:

 - Let's say there is a task called "Multi-assign 2" that has three groups and all three are in the Pending state -- so far so good
Screen Shot 2023-01-31 at 2 16 50 PM

 - Next, let's say that Group 1 starts the task and it becomes Active
 - Although the task is now appropriately displayed in the Active list for Group 1, Group 2 and 3 are no longer displayed in the Pending list

- [x] Can we ensure that Group 2 and 3 would still be displayed in the Pending list?
Screen Shot 2023-01-31 at 2 19 29 PM

 - Next, if the group 1 task is cancelled, it is successfully displayed in the Completed list but a blank Group list is displayed in the Active list

- [x] Can we have Group 1 and Group 2 still remain displayed in the Pending list and have the task removed from the Active list in this instance?
Screen Shot 2023-01-31 at 2 21 12 PM

LunarFang416:
@JoelWiebe Made the small changes you requested.

JoelWiebe (approved):

backend/src/api/workflows.ts
```diff
@@ -338,4 +338,16 @@ router.post('/task/groupTask/:groupTaskID/complete', async (req, res) => {
   res.status(200).json(updatedGroupTask);
 });
 
+/**
+ * Mark group task as active.
+ */
+router.post('/task/groupTask/:groupTaskID/active', async (req, res) => {
```
Ammar-T, line 344: Do you think we should update the above endpoint from `/task/groupTask/:groupTaskID/complete` to `/task/groupTask/:groupTaskID/status` and then pass in [complete, active] in the body? So this would update the status from a single endpoint rather than having different ones for each status.

frontend/src/app/components/ck-monitor/ck-monitor.component.ts
```diff
@@ -260,44 +313,86 @@ export class CkMonitorComponent implements OnInit, OnDestroy {
...
     const groupTasksTableFormat: MonitorData[] = [];
     if (groupTasks) {
       for (let i = 0; i < groupTasks.length; i++) {
-        const groupMembers: string[] = [];
-        for (let j = 0; j < groupTasks[i].group.members.length; j++) {
-          groupMembers.push(
-            (await this.userService.getOneById(groupTasks[i].group.members[j]))
-              .username
-          );
+        if (groupTasks[i].groupTask.status == status) {
```
Ammar-T, line 337: I think this looks fine, but was wondering if we could just embed the workspace angular component into the ck monitor? This would lower the duplicate code right? Or is this different from the workspace?

frontend/src/app/models/workflow.ts
```diff
@@ -84,7 +84,7 @@ export class DistributionWorkflow extends Workflow {
 
 export class TaskWorkflow extends Workflow {
   prompt: string;
-
+  // complete?: boolean;
```
Ammar-T, line 87: I think just remove line if commented.

PR #493, opened 2023-02-06
fix: [477] correctly map required fields

### LINKED ISSUE #477: Distribution Workflow required field bug
### Description
Creating a Distribution Workflow requires a value in the "Randomly Distribute: Posts per Destination" field even if "Copy all with specified tag" or "Copy all with min # upvotes" are selected. 
Screen Shot 2023-01-23 at 6 27 31 PM

**To reproduce bug**
1. Click the "Manage Buckets & Workflows" icon in the nav bar
2. Select the "Create Workflow" tab
3. Select "Distribution Workflow"
4. Fill in all the fields to create a Task Distribution for the "Copy all with specified tag" or "Copy all with min # upvotes" options
5. You will notice that the "Create Distribution Workflow" button becomes enabled, but clicking it fails to create the workflow
6. To create the workflow, you must select the "Randomly Distribute" option, then enter a value into the Posts per Destination field
7. Now you can go back to "Copy all with specified tag" or "Copy all with min # upvotes" and create those workflows

To fix, when "Copy all with specified tag" or "Copy all with min # upvotes" are selected the "Posts per Destination" field should not be required

DCH25:
## Details

- post per buckets no longer required for other non-random workflow distributions

Closes #477

### DIFF
--- a/frontend/src/app/components/create-workflow-modal/create-workflow-modal.component.ts
+++ b/frontend/src/app/components/create-workflow-modal/create-workflow-modal.component.ts
@@ -198,7 +198,7 @@ export class CreateWorkflowModalComponent implements OnInit {
   }
 
   createDistributionWorkflow(): void {
-    if (!this._ppbSelected()) return;
+    if (!this._distributionWorkflowTypeSelected()) return;
 
     const workflow: DistributionWorkflow = this._assembleDistributionWorkflow();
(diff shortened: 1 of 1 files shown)


JoelWiebe (approved):
Thanks, LGTM

Feel free to merge!

Ammar-T (approved):

PR #494, opened 2023-02-08
ISSUE-476: Change workspace actions to HTTP from websockets

### LINKED ISSUE #476: Task Workspace submit button bug
Students cannot submit work if their web socket connection is interrupted

To reproduce: 

1. Open the workspace containing an active task
2. Disconnect from wifi
3. Reconnect to wifi
4. Create a comment in the task workspace

Ammar-T:
## Details

- To avoid the issues caused by websockets when working in the workspace components, we're switching to all the updates being done through HTTP, to improve reliability

Closes #476

### DIFF
--- a/backend/src/agents/workflow.agent.ts
+++ b/backend/src/agents/workflow.agent.ts
@@ -278,55 +278,6 @@ class WorkflowManager {
       return removePostFromSource(source, sourcePosts);
     }
   }
-
-  async updateTask(
-    userId: string,
-    postId: string,
-    type: TaskActionType,
-    delta: number
-  ) {
-    const rawTasks: GroupTaskModel[] = await dalGroupTask.getByUserAndPost(
-      userId,
-      postId
-    );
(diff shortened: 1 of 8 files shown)


JoelWiebe (approved):
Amazing work! This not only fixes the issues of the Submit tag not being enabled when the socket connection is broken but there was another strange bug where the submit tag was only being enabled for the first two or three posts in a tagging task before the submit button stopped being enabled.

I checked the trace logs issue when the socket connection is broken for tags, comments, upvotes, and creating new posts from the board, and the logs are not created when I restart the server backend just before performing the action. 

Can we create another PR for changing tagging, commenting, upvoting, and creating new posts to http requests for all remaining contexts: boards, buckets, list view, etc.?

Once again, awesome work. We'll have to try this again in a busy class--that will be the true test ;) But this is feeling much more robust already.

LunarFang416 (approved):

JoelWiebe:
@Ammar-T Let's merge this one in!

PR #504, opened 2023-03-01
Issue 423 ck monitor todo improvements

### LINKED ISSUE #423: CK Monitor - TODO Monitor improvements
This issue is to make improvements to this TODO Monitor, including the following suggestions:

1. Add "Set as Default" button 

- [x] Add "Set as Default button to the date picker
Screen Shot 2022-12-06 at 1 30 42 PM

- [x] Pressing the "Set as Default" button will (a) save this value for all users and (b) immediately apply this filter for the current user
- [x] Once set, this date range is saved for all users and displays the set range when any user opens the page

2. The "Apply" button can continue to be used as it is now 

- [x] The Apply button should temporarily override the displayed date range for the current user but leave the saved default values unchanged

3. Add cell padding to the table cells (currently padding is set to zero)
Screen Shot 2022-12-06 at 3 08 59 PM

4. Can we show either TODOs or Tasks (not both)? 

- [x] When "Monitor Task" is clicked, hide TODOs
- [x] When "View by TODOs" is clicked hide all task monitors

5. @Ammar-T had mentioned that following looks to be fairly expensive: we'd be making a call for each Todo. 
[frontend/src/app/components/ck-monitor/ck-monitor.component.ts](https://github.com/encorelab/ck-board/pull/415/files/7899094f026d3aad655dd6c46b2c09a54fdeed99#diff-7bc3fca7cb702302ed458d6914de5b834d0a0378b4ef96241e2a4078d23e2a85)
      const date = new Date(`${item.deadline.date} ${item.deadline.time}`);
      const formattedDate = date.toLocaleDateString('en-CA');
      const currentDate = new Date();
      const name = await this.userService.getOneById(item.userID);
It was suggested to be able to pass in an optional parameter through the `this.todoItemService.getByProject(projectID)` request which would return a more "expanded" version of each todo where the username is already loaded. We've done this with a couple things already, here's an example: https://github.com/encorelab/ck-board/blob/develop/backend/src/api/workflows.ts#L223

6. Add values to empty columns

- [x] For the Type column, add the type tags (including coloured backgrounds as used in the TODO tool, see image below)
Screen Shot 2022-12-28 at 10 23 28 PM

- [x] For the Rubric Score column, rename this column to "Completion Quality" and use the completion quality values added from this PR https://github.com/encorelab/ck-board/pull/454

DCH25:
## Details

- Update ck monitor todo table with new fields
- Set default date range to todo table

Closes #423


JoelWiebe (approved):
LGTM!

JoelWiebe:
@DCH25 Let's go ahead and merge this one in! Would you mind resolving conflicts and performing the merge?

PR #320, opened 2022-07-09
Issue 119 enable score authentication

geoffreykwan:
Note: Make sure to test this with the appropriate branches in the other projects such as

- https://github.com/encorelab/SCORE-Docker-Dev/pull/1
- https://github.com/encorelab/SCORE-API/pull/6
- https://github.com/encorelab/SCORE-Client/pull/20

Note 2: Make sure to add these properties to your /backend/.env file. The SCORE_SSO_SECRET can be anything as long as it matches the ck_board_sso_secret_key in your application-dockerdev.properties in SCORE-API.

```
SCORE_SSO_ENDPOINT=http://localhost:80/sso/ckboard
SCORE_SSO_SECRET=VV5jf6rXt4zrbBuH
```
Note 3: Start up CK Board frontend on something besides 4200 because SCORE-Client uses 4200. I've been using 4201 so I've been starting up CK Board frontend like this
```ng serve --port=4201```

## Details

- Implemented SSO authentication with SCORE
- Automatically create teacher and student CK Board accounts if they do not already exist
- Automatically add a student as a member of a project if they try to access a project or board

## Test

- Create a CK Board and add it to a SCORE run
  - Access CK Board and sign in as a SCORE teacher
  - Create a project
  - Create a board
  - Copy the url to the CK Board
  - Go to SCORE and open the Authoring Tool for a run (create a run if you don't have one yet)
  - Add an Embedded step to the project
  - Paste the CK Board url (like http://localhost:4201/project/e347aab0-6623-4ffb-8950-191192956cac/board/eaee497d-3730-451f-a939-a4e5416a9d96) into the "Model File Name" field

- As a student, access the run that contains the CK Board
  - Go to the step that contains the Embedded step that contains the CK Board (It will likely show that you are signed in as the previous user that was already signed in to CK Board. Logout of CK Board inside of the iframe which should then sign you back in to the correct student CK Board user within the iframe. This is an issue we should address later.)
  - If the student does not already have a CK Board account it should automatically create one
  - If the student is not a member of the CK Board project it should automatically add them (If you don't see the board, go to another step in the SCORE project and then come back to the step that contains the CK Board and then you should be able to see the board)

## Things to possibly address in the future
  - When a user signs out of SCORE, we should also sign them out of CK Board. I looked in to this but didn't have an elegant solution for it.
  - Sometimes the student CK Board user is not automatically added to a project when they should be

Closes #119

PR #508, opened 2023-03-04
(FIX): Redirect to initial URL upon login

### LINKED ISSUE #505: Redirect to initial URL upon login
### Description
If someone is logged into SCORE (but not CK Board) and navigates to a step that makes a request to a CK Board embedded URL, then ensure the user is sent to that destination CK Board embedded URL rather than the project dashboard

**Problem**
In testing the CK Board “embedded URLs” (e.g., https://ck-board.oise.utoronto.ca/project/29c3a9f5-459c-455d-bad0-914a160d7a9c/board/ee2c5e87-de1c-4bf7-8105-c723413249bb?embedded=true) I had always opened the CK Board in its own tab before opening up a board in SCORE. I just realized that if I am logged out of CK Board when I navigate to an embedded CK Board URL then I get booted to the CK Board dashboard and the “embedded” URL parameter is removed.

I believe this is the result of the first request to CK Board triggering the sign in that redirects to the dashboard. 

This figure shows the user being re-routed back to the dashboard on initial access
Screenshot 2023-03-03 at 11 47 40 AM

This figure shows the proper display of the CK Board canvas after refreshing the browser within SCORE
Screenshot 2023-03-03 at 11 48 26 AM

LunarFang416:
## Details

- If a user is not logged into CK-Board and tries to navigate to any url (including embedded urls), they will be redirected to the login page, and after they login, they will be redirected to the URL they originally wanted to go to (if they are authorized to)
- Tested for all embeded URL's (todo, ck-monitor, ck-workspace, board)

- Weird bug for board embeddable URLs was occured when it was running the AuthGuard, it would simultaneously run the BoardGoard and ProjectGuard as well. The AuthGuard would redirect to /login momentarily, but would have a race condition with the BoardGuard, which would then send it to /error since you aren't logged in and can't access API. To quickly remedy this, i added a if statement in BoardGuard and ProjectGuard, to see if the user is logged in, if not it would return false, and the AuthGuard would reroute to /login properly.

Bug:
- To redirect to URL with queryParams, we need to parse them and send it as a separate object to `this.router.navigate`, which it didn't do before.

@JoelWiebe  Could you test this out and see if it works as we want when used in SCORE? It works fine when using it just with CK-Board, so theoretically it should do in SCORE as well.

Closes #505

### DIFF
--- a/backend/src/utils/auth.ts
+++ b/backend/src/utils/auth.ts
@@ -109,7 +109,9 @@ export const getParamMap = (params: string): Map<string, string> => {
   const paramMap = new Map<string, string>();
   const keyValuePairs = params.split('&');
   keyValuePairs.forEach((keyValuePair) => {
-    const [key, value] = keyValuePair.split('=');
+    // Get everything before the = as the key and everything after the = as the value
+    const splitResult = keyValuePair.split(/=(.*)/);
+    const [key, value] = splitResult;
     paramMap.set(key, value);
   });
   return paramMap;
--- a/frontend/src/app/components/login/login.component.ts
+++ b/frontend/src/app/components/login/login.component.ts
@@ -17,7 +17,9 @@ export class LoginComponent {
   invalidCredentials = false;
 
   constructor(private userService: UserService, private router: Router) {
-    if (this.userService.loggedIn) this.router.navigate(['/dashboard']);
+    if (this.userService.loggedIn) {
+      this.router.navigate(['/dashboard']);
+    }
   }
 
   onLogin() {
@@ -28,7 +30,21 @@ export class LoginComponent {
         const redirectUrl = this.userService.redirectUrl;
         this.userService.redirectUrl = null;
         if (redirectUrl) {
-          this.router.navigate([redirectUrl]);
+          // Parse query parameters (if they exist)
+          const searchParams =
+            redirectUrl.indexOf('?') != -1
+              ? new URLSearchParams(
+                  redirectUrl.substring(redirectUrl.indexOf('?'))
+                )
+              : [];
+
+          this.router.navigate([`${redirectUrl.split('?')[0]}`], {
+            queryParams: [...searchParams].reduce((o, [key, value]) => {
+              o[key] = value;
+              return o;
+            }, {}),
+            queryParamsHandling: 'merge',
+          });
         } else {
           this.router.navigate(['dashboard']);
         }
--- a/frontend/src/app/guards/auth.guard.ts
+++ b/frontend/src/app/guards/auth.guard.ts
@@ -17,15 +17,20 @@ export class AuthGuard implements CanActivate {
     public ngZone: NgZone
   ) {}
 
-  canActivate(_next: ActivatedRouteSnapshot, _state: RouterStateSnapshot) {
+  async canActivate(
+    _next: ActivatedRouteSnapshot,
+    _state: RouterStateSnapshot
+  ) {
     if (this.userService.loggedIn) {
       return true;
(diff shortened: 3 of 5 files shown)


JoelWiebe (changes requested):
Thanks for your quick work on this!

- Redirecting to an embedded URL now works great when prompted to login with CK Board
- For some reason, when prompted to log in with SCORE we are still sent to the dashboard

@geoffreykwan had mentioned that SCORE sends the following to CK Board when redirecting with the embedded parameter:

nonce=873b9aaa-878e-429c-9473-c43316a60b78&user-id=1254&username=aa0101&role=student&redirect-url=/project/d5d5c173-8e24-443a-885b-9c953e4e3eea/board/f83f886f-edbe-4998-98fe-e21131b94923?embedded=true

LunarFang416:
@JoelWiebe It must be failing to redirect to the original url since `this.userService.redirectUrl` is null after the SSO is successful, because if `this.userService.redirectUrl` is null it redirects to the dashboard, which is happening now.

I don't have SCORE set up right now, so I won't be able to to see how it handles the embeddable url when requested from SCORE.

The backend API does look like it is parsing the `redirect-url` query paramter from the URL, maybe we aren't setting `this.userService.redirectUrl` when the response is returned.

Don't want to create a commit for this, but I think we can try to see if something when we add this line, which will set the `this.userService.redirectUrl` when SSO is successful. This is in the `user.service.ts` file:
```typescript
  async ssoLogin(sso: string | null, sig: string | null): Promise<boolean> {
    return this.http
      .get(`auth/sso/login/${sso}/${sig}`)
      .toPromise()
      .then((result: any) => {
        localStorage.setItem('user', JSON.stringify(result.user));
        localStorage.setItem('access_token', result.token);
        localStorage.setItem('expires_at', result.expiresAt);
        window.location.href = result.redirectUrl;
        this.redirectUrl = result.redirectUrl; // Addition; sets this.userService.redirectUrl after SSO successful
        return true;
      });
  }
```
 @JoelWiebe Could you add this line and see if something changes? 
Not fully confident since I don't have SCORE set up, so might need some help with this

JoelWiebe:
@LunarFang416 Thanks for the feedback and the suggestion!

Somehow after setting that line, I'm still getting the boot to the project dashboard.

JoelWiebe:
P.S. @LunarFang416 Here are the file changes that support CK Board authentication: https://github.com/encorelab/ck-board/pull/320/files

If you cannot get this working, no stress, I'll find a workaround for Monday.

geoffreykwan:
I think I fixed the redirect problem not keeping embedded=true. The way we were parsing the params was splitting on the = character and getting the first value and second value from the split but it doesn't work correctly if there is more than one = character. What we need to do is split on the first = and then get everything before the = as the first value and then everything after the = as the second value.

Before we were ending up with this redirect url (notice embedded at the end does not include =true)
`/project/88d328f8-b5c0-4c3d-95b1-8866226a1aa8/board/c6e30415-6819-4d98-baa1-27f67178f138?embedded`

Now with the fix we end up with this redirect url
`/project/88d328f8-b5c0-4c3d-95b1-8866226a1aa8/board/c6e30415-6819-4d98-baa1-27f67178f138?embedded=true`

I also needed to add a check in the AuthGuard to not redirect to `/login` when SSO is enabled.

JoelWiebe (approved):
Brilliant! Thank you both for your work @LunarFang416 @geoffreykwan 

LGTM

JoelWiebe:
@LunarFang416 we should be good to merge!

PR #511, opened 2023-03-15
ISSUE-509: Disable button while post is being created

### LINKED ISSUE #509: Prevent duplicate new posts and updates
### Description
With a number of concurrent users, the Add Post modal occasionally remains open after the Add Post button is clicked allowing users to generate many duplicate posts 

**Details**
For a few students during an activity, the new post modal stayed stuck open until the student pressed it 25 times then it created all 25 posts on the board. This would imply that some kind of error had occurred during the creation of the post preventing the modal from closing (i.e., maybe an issue with web sockets or some other problem resulting from many concurrent users or DB updates). Do we think the switch to HTTP requests instead of web sockets would fix this or is there another fix that we should think about to ensure an opened modal can only be submitted one time?

**Task**

- [x] When the Add Post button is clicked, save the title and message to local storage then close the modal immediately
- [x] If the post was successfully saved, then do not display any message and delete the title and message from local storage
- [x] If an error occurred, open a snackbar with the message: "Unable to create post, please try again!"
- [x] If an error had occurred on the previous Add Post submission, then, when the Add Post modal is opened, populate the title and message fields with the saved values
- [x] If no error occurred on the previous Add Post submission, then, when the Add Post modal is opened, the title and message fields should be blank (as per usual)

Ammar-T:
## Details

- When post is bring created, the "create post" button will be disabled until completion
- Added snackbars in case of error or if post is successfully created

Closes #509

### DIFF
--- a/frontend/src/app/components/add-post-modal/add-post.component.html
+++ b/frontend/src/app/components/add-post-modal/add-post.component.html
@@ -193,7 +193,7 @@
   <button
     *ngIf="!editingPost"
     type="submit"
-    [disabled]="disabled()"
+    [disabled]="disabled() || this.creationInProgress"
     mat-button
     (click)="handleDialogSubmit()"
     cdkFocusInitial
--- a/frontend/src/app/components/add-post-modal/add-post.component.html
+++ b/frontend/src/app/components/add-post-modal/add-post.component.html
@@ -193,6 +193,7 @@
   <button
     *ngIf="!editingPost"
     type="submit"
+    [class.spinner]="this.creationInProgress"
     [disabled]="disabled() || this.creationInProgress"
     mat-button
     (click)="handleDialogSubmit()"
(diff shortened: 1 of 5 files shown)


JoelWiebe:
@Ammar-T Thanks for you work on this!

You've created a standard that we should probably apply across the board (as another issue) for all of our modals (e.g., settings), e.g., disabling buttons and adding the spinner, then providing a snackbar error message. I was able to see the button spinner when adding 2 second network latency in the browser. 

Just two very small suggestions:

- [ ] Can we adjust the snackbar text to: "Unable to create post. Please refresh and try again!"
- [ ] Also, the line spacing seems very large. Can we reduce the line spacing to keep the message more contact, like we have for the task description? (see below)

Screenshot 2023-03-23 at 5 25 15 PM
Screenshot 2023-03-23 at 5 28 01 PM

Ammar-T:
@JoelWiebe Changed the error message's text, but the spacing issue would be better handled in a separate issue.

PR #512, opened 2023-03-18
(FIX): Posts can be downloaded from List and Bucket View error-free

### LINKED ISSUE #507: Fix downloading post from list view and bucket
### Description
Attempting to download a post from the list view or bucket throws an error and the post is not downloaded onto the canvas.

**Re-produce bug**

1. Open either the list view or an existing bucket
2. Click the Create Post icon to create a new post
3. You should see a download icon on that post in the list view or bucket, click the download icon
4. An error appears in the console and the post is NOT downloaded to the canvas

LunarFang416:
## Details

- Previously we were trying to update posts by passing the flattened version of the post to the `$set` operator, but the `$set` operation doesnt work with dot notation, so this would be the cause for error:

    ```typescript
    {
      '$set': {
        type: 'BOARD',
        title: 'testing bug',
        desc: '',
        tags: [],
        multipleChoice: undefined,
        'displayAttributes.position.left': 720,
        'displayAttributes.position.top': 219,
        'displayAttributes.lock': false
      }
    }
    ```
- If we simply pass the fields of the Post that we want to modify to the `$set` operator directly, it will update them successfully, and the post will be downloaded
- Tested for both List and Bucket Views

Closes #507

### DIFF
--- a/backend/src/repository/dalPost.ts
+++ b/backend/src/repository/dalPost.ts
@@ -97,11 +97,11 @@ export const removeByBoard = async (boardID: string) => {
 
 export const update = async (id: string, post: Partial<PostModel>) => {
   try {
-    const setPost: any = flatten(post);
-
-    const updatedPost = await Post.findOneAndUpdate({ postID: id }, setPost, {
-      new: true,
-    });
+    const updatedPost = await Post.findOneAndUpdate(
+      { postID: id },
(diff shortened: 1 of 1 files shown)


JoelWiebe (approved):
LGTM

Let's go ahead and merge!

Issue #484, opened 2023-01-24
Separate "Unassigned" group member list container


JoelWiebe:
### Description
Separate "Unassigned" group member list into its own column on the right side

**Problem**
When dealing with many groups with a large number of students (e.g., 6 groups with 5 students each), the unassigned column gets pushed way down to the bottom making it impossible to drag unassigned students to the top group containers (as they are out of view)
<img width="1305" alt="Screen Shot 2023-01-23 at 9 49 33 PM" src="https://user-images.githubusercontent.com/6416247/214202505-87f8680d-fbd8-426c-b6a4-93da509f603d.png">

**Details**

- [ ] Move the "Unassigned" group members list container in a fixed location on the right side
- [ ] Allow for scrolling up and down through the grid of available group member list containers in the left pane

PR #516, opened 2023-03-29
feat [342]: update group modal UI flow

### LINKED ISSUE #342: Grouping UI - Touchups
1. Ensure the Back and Update buttons remains visible (currently they can scroll out of view; see image below)
Screen Shot 2022-12-23 at 7 32 17 PM

 - you can use the board configurations as an example of these buttons remaining in view while contents scroll
Screen Shot 2022-12-23 at 7 33 07 PM

2. Like our other modal pop-ups, add a Close button to the **bottom** of our Groups tab (see group tab currently without "Close" button below)
Screen Shot 2022-12-23 at 7 36 20 PM

4. For the multi-assign tab, (A) similar to above, ensure the buttom buttons don't scroll out of view but also (B) add a "Close without saving" button (as seen below)
Screen Shot 2022-12-23 at 7 38 13 PM

5. If the teacher (i) makes a change and (ii) clicks outside of the modal pop-up (which signals to close the pop-up), then create a confirmation pop-up (as seen below)-- message: "You have unsaved changes"; buttons "Save groups" & "Close without saving" [UPDATE: The image below should say "Close without saving" NOT "Cancel without saving"
Screen Shot 2022-12-23 at 7 48 42 PM

6. If either of the "Save groups" buttons are clicked (in the pop-up or in the Multi-assign tab), could we keep the same group columns open and display a snackbar message: "Successfully saved groups")?

DCH25:
## Details

- Added confirmation modals and UI improvements

Closes #342

### DIFF
--- a/frontend/src/app/components/confirm-modal/confirm-modal.component.html
+++ b/frontend/src/app/components/confirm-modal/confirm-modal.component.html
@@ -1,6 +1,6 @@
 <h1 mat-dialog-title>{{ title }}</h1>
 <div mat-dialog-content>{{ message }}</div>
 <div mat-dialog-actions>
-  <button mat-button mat-dialog-close (click)="handleConfirm()">Confirm</button>
-  <button mat-button mat-dialog-close (click)="handleCancel()">Cancel</button>
+  <button mat-button mat-dialog-close (click)="handleConfirm()">{{ data.confirmLabel? data.confirmLabel : 'Confirm' }}</button>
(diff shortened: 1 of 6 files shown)


JoelWiebe (commented):
@DCH25 Wonderful work! This is great.

- [ ] Could you make one very small change? As per item 5 in the issue, UPDATE: The text for the modal should say "Close without saving" not "Cancel without saving"

- [ ] Can you also merge with develop and ensure that the unassigned column remains stationary on the right side, as per https://github.com/encorelab/ck-board/issues/484

Once both of these are complete, then please feel free to merge!

PR #518, opened 2023-03-31
issue 488 send to login screen on session expiration

### LINKED ISSUE #488: Send to login screen on session expiration
Send users to the login screen if the session expires

vkorir:
Set a timeout that will redirect the user to the login page once the access_token expires. The timeout time is set after successful login and period is equal the time until access token expires.

Closes #488


JoelWiebe:
@vkorir Great work! I just tested this out. As our production server uses SCORE SSO, we should probably try getting the SCORE SSO cases working before merging these changes into develop. The current draft PR has the following effect:

1. For standalone CK Board, everything looks good (redirected to the login page when session expires)
2. With SCORE SSO, the user is not redirected to login again when the session expires but is still prompted to login when navigating between "pages" once the session is expired (i.e., from project dashboard to board dashboard), however:
 a. if the SCORE session is still active when the CK board session expires, then the CK Board will simply be routed to the project dashboard (the user does not need to login again)
 b. if the SCORE session has ended when the CK board session expires (e.g., user logged out of SCORE) then CK Board will be redirect to the SCORE SSO page

Suggestions:
- [ ] Are you up for trying to get CK Board to route the user to the login page when the session expires while using SCORE SSO?
- [ ] If SCORE is still active and the CK Board session has expired, is it possible for the user to not be redirected back to the project dashboard during renewal of the session but be redirected to stay at the current URL with existing parameters? This becomes important if we are embedding URLs to particular CK Board canvases in SCORE

vkorir:
I was looking at SCORE authentication flow, and seems like we use HttpOnly cookies. This means it's inaccessible by JavaScript and there's not really a way to poll until when the session expires to redirect users. @geoffreykwan, can you confirm this?

geoffreykwan:
Yea we have HttpOnly cookies set to true for security purposes. Are you trying to handle the scenario where CK Board wants to check if the SCORE session is still alive?

I just re-read through the comments in this PR https://github.com/encorelab/ck-board/pull/412 and I think we were able to get the SCORE session token string sent to CK Board and saved for later use. This way when a user logs out of CK Board, CK Board can also send a logout request to SCORE to tell SCORE to log out. I think you should be able to use this same process to have CK Board make a request to SCORE to check if the SCORE session is still alive.

See the function here
https://github.com/encorelab/ck-board/blob/0a5dd9470ca82afc9f5e524259d32c56790e4ccf/backend/src/utils/auth.ts#L244-L252

You can check if a user is still logged into SCORE by making a request to `https://score.oise.utoronto.ca/api/user/info`

If the SCORE user is logged in, you will get a response that looks like this
`{"schoolLevel":"COLLEGE","lastName":"kwan","isPreviousAdmin":false,"country":"USA","role":"teacher","isGoogleUser":false,"city":"New York","displayName":"geoffrey kwan","language":"en","firstName":"geoffrey","id":1181,"state":"New York","schoolName":"University of California Berkeley","email":"","username":"geoffreykwan"}`

If the SCORE user is logged out, you will get a response that looks like this
`{"username":null}`

PR #523, opened 2023-04-05
ISSUE-515: Multi-Model

### LINKED ISSUE #515: Multi-Model Learner Model (Part II)
Begin updating the current learner model setup to allow for multiple types of learner models. Currently, we only have the engagement-based model, but we want to branch out towards "SEL" and "Content" models as well.

1. Update the UI to remove buttons/text similar to "View Engagement Model" and add something more generic like "View Learner Models". Once clicked, create some way of filtering or viewing all of the models, not just the engagement model.
2. Add ability to import/export data to and from the created learner models via csv files. File formatting will be open to the assignee and should be implemented to create an efficient mapping from learner model format in the DB to the csv formatting.
3. Add pop-up warning: “Do you want to overwrite all existing data with imported values for this data model?” “Overwrite all” “Cancel”.

Ammar-T:
## Details

- [x] Allow creation of multiple custom learner models
- [x] Create default models for engagement, SEL, and content
- [x] Add import/export feature for models

Closes #515


Ammar-T:
@JoelWiebe Added all the requirements from #515 except the last optional one.

Based on yesterday's discussions regarding including project id and other fields in the csv, I've made the export file to include the project and model name in the file name to make them easy to distinguish. The actual content of the file do not include those fields. 

I also added a "Help" section where users would upload the csv file to ensure they follow the required format and have access to student ids as well.

Only thing remaining for this is some error-handling. For example, if the csv is improperly formatted we need to tell the user.

JoelWiebe:
@Ammar-T Thanks for your work on this; really great work! 

Tested CSV validation for: 

- Incorrect student ID
- Diagnostic/reassessment values over 100

I have a collection of small requests/ideas: 
1. When downloading the student info, can we add the dimension, diagnostic, and reassessment headers to the CSV (to make it really easy to use this as a template for importing)?
2. I didn't see the project name and ID in the CSV, but if it's not there that is okay (just adds a little extra context info)
3. When clicking the Clear button to delete all data from the model, can we provide a warning dialogue (e.g., Do you want to permanently delete all data from this model? Cancel; Delete)
4. Can we add a line, something like the following, to the help tool: "Diagnostic and reassessment values must be between 0 and 100 (inclusive)"?
5. Do we still have an individual student view of all models? For instance, can we add each student's name to the top-right drop down list--when selected it could display that particular student's diagnostic, reassessment, class avg diagnostic, and class avg reassessment for each model? Was this scheduled for a third PR?
6. Can we put the Show/Hide Learner Models button on its own row?

Sorry for all the extra items! Most are pretty quick fixes. This will probably be your last PR of April though :)

Ammar-T:
@JoelWiebe 

1. - [x]  When downloading the student info, can we add the dimension, diagnostic, and reassessment headers to the CSV 
2. I added the project name to the csv's file name to make it clearer to the teacher. I feel adding it to the file contents wouldn't be as effective since it's not used in the application logic (i.e. when parsing the csv).
3. This is already being done? When clicking "Clear Data" there's a warning modal.
4. - [x] Added line to help tool
5. This is already in the PR as well. Are you sure you've pulled all the recent changes?
6. - [x] Moved buttons to their own rows

JoelWiebe (changes requested):
@Ammar-T 

I'm really impressed by your work on this!

Just a few very small items/suggestions:

1. It appears that the charts are mislabeled with a max value of 50 rather than 100
2. Can we change "Export via CSV" to "Export as CSV"
3. When creating a new Learner Model, can we have the Model Name on the same tab as the dimensions and data (as opposed to Information and Data tabs; it seems a little more intuitive if all together in one space, especially if they start to create dimensions then don't know that the model name is a required field to submit)?

Functionally, everything seems to work quite well -- and visually this looks sharp!

JoelWiebe (approved):
To get this feature in for tomorrow's run, let's fix these last few items in a following PR.

Ammar-T:
1. I think the 50 is just showing the halfway point of the axis. So although a bit confusing, I think the value is ok.

Screen Shot 2023-06-01 at 9 54 16 PM

2. Changed!
3. Makes sense, changed it!

Changes in #543

PR #528, opened 2023-04-22
ISSUE-521: Save to personal board posts centered

### LINKED ISSUE #521: Fix "Save to Personal Board"
### Description
A small update and performance check are needed for the "Save post to personal board" feature

Screen Shot 2023-04-04 at 5 13 32 PM

**Details**
Currently when the post is saved to the personal board, it's copied using the Community Board location -- instead we would prefer that it be dropped in the center of the default view of the personal board (similar to downloading a post from a bucket; see https://github.com/encorelab/ck-board/pull/162 and https://github.com/encorelab/ck-board/pull/512.

Additionally, some people thought that their post might not have been saving during a class run. It would be helpful to verify whether this uses http requests or web sockets.

**Tasks**

- [x] Send Starred post to the centre of the personal board (not some other position)
- [x] Verify that this Star feature uses http requests rather than web sockets; if using web sockets, switch to http requests

DCH25:
## Details

- Personal board posts created from "Save to personal board" button are centered by default
~~- Switch to only use http~~
  ~~- Remove associated websocket/trace implementations~~

Closes #521


JoelWiebe (commented):
Thanks for your quick work on this!

Would it be possible to have each of the downloaded posts to have a slight offset? For buckets, for example, when a [bucket is opened and the download icon is clicked](https://github.com/encorelab/ck-board/pull/162) for multiple posts each of the downloaded posts will be slightly offset from one another (so that they are each visible on the board). It would be great if we could do something like that for saving posts to the personal board.

JoelWiebe:
@DCH25 can you finish this one? It should be pretty quick!

**Details**
With buckets, once you open a bucket view and download two posts the first will land in the center of the canvas and the second will be slightly offset below the first. Once you close the bucket view then open it again, the first downloaded post will again land perfectly in the center then the second will again be slightly vertically offset.

Downloading posts from the bucket view to the canvas (New Post 1, New Post 2, then New Post 3)
Screenshot 2023-07-05 at 10 47 41 AM

Posts are displayed with a vertical offset on the canvas
Screenshot 2023-07-05 at 10 50 43 AM

Can we do the same for Save Posts to Personal Board?

**Tasks**
- [ ] When a canvas is first opened and a post is saved to the personal board, it is saved to the center of the personal board (as it currently does)
- [ ] When a canvas remains open and a second post is saved to the personal board, it is saved with a vertical offset from the center of the personal board 
- [ ] When the user navigates away from the canvas (e.g., to the dashboard or another canvas) then back to the canvas and saves a post to the personal board, the post is again saved perfectly to the center of the personal board

JoelWiebe (approved):
Thanks, LGTM

PR #535, opened 2023-05-20
(FEAT): Improve Generative task workflow logic and updates to workspace UI

### LINKED ISSUE #534: Workspace UI upgrades
Workspace
We will be making some changes to the workspace, these changes will apply to both Peer Review and Generation tasks

image
- [ ] For both peer review and generation, move the title to group and workflow name to the top of the page (e.g., Group 3: SEL Challenges Brainstorm)
- [ ] For both peer review and generation, move the prompt just below the group and workflow name (e.g., What are four challenges...)
- [ ] For both, move the requirements just below the prompt as an unordered list (e.g., Create 4 posts; Tag each post at least once)
- [ ] For only the generation workflow, add a plus button that opens the Add Post modal. Submitting the post (1) adds it to the swiper tool and (2) increments the progress bars. 
- [ ] Generation workflow posts follow the same flow as Peer Review, moving from Pending, to Active, to Completed
- [ ] For both peer review and generation, once all requirements are met, the status is automatically set "Complete" (the "Mark as Complete" button can be removed as this is no longer needed
- [ ] For both peer review and generation, add a toggle button for View Draft & View Submitted that swaps the list of posts in the swiper from those that have yet to be submitted and those already submitted. This toggle should be defaulted to show the Drafts.
Screenshot 2023-03-29 at 3 37 40 PM

LunarFang416:
## Details

- Updates made to CK Workspace UI
- Changes to Generative workflow:

     - When a user creates a post, it will be added to the drafts
     - Once all requirements are met, the "Submit" button will be enabled. When submit is pressed the post will then be transferred to the respective bucket/board destination

Closes #534


JoelWiebe:
@LunarFang416 Awesome work! Could we make some slight styling changes:

- Change the prompt to a slightly smaller font, perhaps h2
- Add a margin above the "Group members" label, perhaps 16px 0 0 0 and set this label to bold

Screenshot 2023-06-29 at 10 15 26 AM

JoelWiebe:
- [ ] Can we add 16px margin between the control (for switching between drafts and submitted) and the div with the cards of posts?
- [ ] Can we change the control from a slide toggle to a single selection button toggle (perhaps with the options View Drafts and View Submitted)?
- [ ] When I click Begin Task to move a generation task from Pending to Active, the plus button becomes skewed; however, when I refresh the browser then open the task it presents just fine

Screenshot 2023-06-29 at 11 12 13 AM

- [ ] For a generation task, when a task is tagged when it is first created, the tag is not counted toward the requirement of tagging the post; I must re-open and re-tag the post for it to be counted (this could be a broader bug were maybe the tag is not being logged in the backend either when tagged at the time of creation -- we had this bug before)
- [ ] For generation tasks, when multiple posts are created and ready to be submitted, when only one of them is submitted, they are all automatically created; the expected behaviour was that each of the posts will require that its Submit button be clicked but perhaps this is too tedious. What do you think about us (1) renaming the View Submitted button to View Completed, (2) removing the Submit button on each post, and (3) as soon as requirements for a post are met, the post is moved from the Drafts View to Completed View?
- [ ] I was going to ask if we can disable the "Submitted!" button (seen below) on the cards for submitted posts (so it is obvious they cannot be clicked), but if we do the change above, this change will no longer be required (as the button will be gone)
Screenshot 2023-06-29 at 10 20 59 AM

- [ ] In the Create Workflow tab, can we rename and re-order the workflow radio buttons (as seen below) to Generation Task, Peer Review, and Distribution with Generation Task as the default selected option? 
Screenshot 2023-06-29 at 12 03 34 PM

JoelWiebe:
@LunarFang416 feedback:

The changes look great, there are just two changes I'm not seeing:

- [ ] Add post button for generative workflow in workspace is stretched when I first click the "Resume Task" or "Begin Task" buttons 
- [ ] There is not 16px margin between the View Drafts button and cards div
Screenshot 2023-07-19 at 10 35 48 AM

Thanks for all your work.

JoelWiebe (approved):

PR #539, opened 2023-05-28
fix: disable todo monitor "Mark as Done" option

DCH25:
## Details

- Todo monitor "Mark as Done" option now properly disables todo completion


JoelWiebe (approved):
LGTM

PR #555, opened 2023-07-12
Close all snackbars after 5 seconds

### LINKED ISSUE #553: Fading Snackbars
### Description
Across the platform, snackbar messages provide feedback to the user but remain open, where all must be manually closed. Instead, we would like each snackbar to close after 5 seconds

**Details**
Snackbars should be able to be closed by configuring its "duration" option.

**Tasks**
- [x] Close all snackbars across CK Board after 5 seconds

vkorir:
- Added snackbar duration config to close all snackbars in 5 seconds

Closes #553


LunarFang416 (approved):
👍

PR #557, opened 2023-07-19
(FEAT): Learner models set to project level

### LINKED ISSUE #551: Set Learner Model to Project Level
### Description
Move learner models to the project level rather than the board level 

**Details**
The same learner models should be accessible from all boards within a CK board project; currently, I appear to need to open a particular board to access the learner model created from the context of that particular board

LunarFang416:
## Details

- Learner Models are now at the project level; they can be viewed from any board and will the same learner models
- Any change in any board is propagated at the project level, so changes are now at the project level rather than the board level

#### For existing projects
- For all projects before this change, the learner models seen in CK-Monitor will have an aggregate of all the learner models created in the boards individually i.e if previously your project has 3 boards each with 3 learner models, this change will show a total of 9 learner models at the project level
- This will preserve any existing learner models and you can delete then at your choice without having to export and repopulate an existing learner model

Closes #551

### DIFF
--- a/backend/src/api/learner.ts
+++ b/backend/src/api/learner.ts
@@ -27,6 +27,13 @@ router.post('/board/many', async (req, res) => {
   res.status(200).json(models);
 });
 
+router.post('/project/many', async (req, res) => {
+  const { projectIDs } = req.body;
+
+  const models = await dalLearnerModel.getByProjects(projectIDs);
+  res.status(200).json(models);
+});
+
 router.post('/:id/addDimension', async (req, res) => {
   const { id } = req.params;
   const { dimension } = req.body;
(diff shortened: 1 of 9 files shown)


JoelWiebe (approved):
Looks great!

Just out of curiosity, do you know which learner models would be preserved? I.e., if an existing project has learner models in one of its boards, do these changes preserve the learner models of the first board in the project or the first set of learner models created in any board? If this is unclear, that's okay -- just curious. I will download the learner models from the one project on the production server then re-upload them after we pull these changes to the server.

LunarFang416:
@JoelWiebe It will preserve all learner models in the project. So if a project has 3 boards and each of them have 3 learner models each, it will now show all of them combined (so a total of nine). So none of the learner models will be lost, they will all  just appear together now.
You won't have to download the learner models since it'll preserve all learner models!

PR #574, opened 2024-03-28
WIP: refactor(backend): Socket state management refactor + adding redis adapter

LunarFang416:
## Details

- This PR attempts to fix the socket management issues we are currently facing
- There were a few errors in the code regarding how we were managing rooms and users, which I fixed
- Did thorough testing for a lot of different flows to make sure events are propogated
- Also added a redis-adapter to the socket server so that when we are running servers on scale, different users connected to the same room but on different servers will also receive events

Closes #{ISSUE-NUMBER}

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.0.0",
       "license": "ISC",
       "dependencies": {
+        "@socket.io/redis-adapter": "^8.1.0",
         "@typegoose/typegoose": "^9.8.1",
         "@types/bcrypt": "^5.0.0",
         "bcrypt": "^5.0.1",
@@ -18,6 +19,7 @@
         "dotenv": "^16.0.0",
         "express": "^4.17.3",
         "express-jwt": "^7.7.0",
+        "ioredis": "^5.3.2",
         "jsonwebtoken": "^8.5.1",
         "migrate-mongo": "^9.0.0",
         "mongo-dot-notation": "^3.1.0",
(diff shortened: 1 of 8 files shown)


JoelWiebe (commented):
Great idea using the redis adapter. Hopefully once we figure out these couple items this will provide us with some much needed stability and reliability. 

Tested:

1. Canvas: Create and tag posts on a canvas with two concurrent users
2. Workflow: Generate 2 posts in canvas **(Bug)**
3. Workflow: Generate 3 posts in bucket
4. Workflow: Move ideas between buckets
5. Workflow: Peer review ideas to move between buckets
6. Workflow: Peer review ideas from canvas to bucket **(Bug)**
7. Workflow: Generate 4 posts to canvas **(Bug)**

Bugs (may not be related to web sockets):
2. 

- [ ] (a) When creating posts in the Workspace, posts created with tags are not displayed on the canvas. 
- [ ] (b) If the final post created has a tag then the task fails to be marked as complete. 
- [ ] (c) Posts are only displayed in the canvas after the canvas is refreshed--should they be showing up immediately?

6. 

- [ ] In the Workspace, when peer reviewing posts from the canvas to a bucket, all posts across the project were displayed as posts to review (including all posts from buckets and the two posts that did not appear on the canvas from the task above). Conclusion: This workflow is mistakenly providing all posts for peer review rather than only posts on the canvas.

7. 

- [ ] In the Workspace, I created all four posts with no tag. Upon completion they were all added to the canvas, but posts 1, 2, and 3 were posted twice and the 4th only once. NOTE: Could not replicate this.

Screenshot 2024-04-10 at 10 25 55 PM

- [ ] Additionally, I'm noticing only canvas events are being logged. No posts tagged or created in the Workspace are being logged. Is this unrelated to web sockets?

LunarFang416:
Will create new issues for the existing bugs, merging this in for now.

PR #641, opened 2024-12-15
635 feature boards permission updates

markiianbabiak:
## Details
Verified that the teacher can access all views and view permission in board settings only applies to students' views; also, disabled student tabs are completely hidden rather than disabled
- 

Closes #{635, 573}


JoelWiebe (approved):
Great work. Looks good to me!

Feel free to merge.

PR #643, opened 2024-12-28
640 bug handle deleting post during generation workflow

### LINKED ISSUE #640: BUG: Handle deleting post during generation workflow
### Bug
**Overview**
If a student deletes a post during a generation task after having created one, the progress bar does not decrease by one and the task can be completed with one post too few. 

**To Reproduce**
Steps to reproduce:
1. Create a generation task with two or more posts to generate
2. Create a post then delete it
3. Create another post
4. You will see that the progress bar suggests that two posts are created. If you create all posts, the task will be marked as completed when total required - 1 are created. 

**Proposed tasks to fix**
- [x] Deleting a post should reduce the progress bar for both the teacher monitor and the student workspace progress bar
- [x] Ensure task is only marked as completed when the total number of generated posts are created (ignoring the deleted posts) -- this behaviour may already exist once the prior item is fixed

**Screenshots**

markiianbabiak:
## Details
If a student deletes a post during a generation task after having created one, the progress bar decreases and the task requires one more post to be completed.
- 

Closes #640

### DIFF
--- a/frontend/src/app/components/ck-workspace/ck-workspace.component.html
+++ b/frontend/src/app/components/ck-workspace/ck-workspace.component.html
@@ -1,75 +1,202 @@
 <div>
-  <app-toolbar class="toolbar" *ngIf="user && project && board && !embedded" [user]="user" style="width: 100%;">
+  <app-toolbar
+    class="toolbar"
+    *ngIf="user && project && board && !embedded"
+    [user]="user"
+    style="width: 100%"
+  >
     <span *ngIf="board">CK Workspace - {{ this.board.name }}</span>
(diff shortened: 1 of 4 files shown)


JoelWiebe (changes requested):
The progress bar is working great and completes the task appropriately. 

I notice that after creating the first post and then immediately deleting it, that the post viewer is still visible.

Screenshot 2025-01-07 at 2 51 43 PM

Then after I recreate the first post, it clips. 

Screenshot 2025-01-07 at 2 53 04 PM

Can we have the post viewer completely removed if the first post is deleted, then ensure subsequent posts display without clipping?

Thanks for getting the progress bar working!

PR #654, opened 2025-01-25
Navigate to Login page on invalid session token

### LINKED ISSUE #651: FEATURE: Session expiration redirect to login page
### Feature
**Overview**
When the session expires, redirect the user to the login page

**Details**
Currently no redirect happens, content simply stops updating with 403 errors, then on refresh the user is navigated to the index.html page rather than the login page

**Tasks**
- [ ] Throw 401 error when the session expires
- [ ] Catch 401 error and redirect to the login page

JoelWiebe:
## Completed

- Extended session duration from 2h to 4h
- Modified API Interceptor to navigate to login page on 401 error
- Modified invalid token error to throw 401 error

## Test

- [ ] Verify no unusual browser console or backend errors (other than the one or two 401 errors thrown immediately at the time a resource is requested when the session token has expired)
- [ ] Ensure navigation is routed to the login page when the session expires and a resource is requested

To test, you may need to rebuild the backend and it will be helpful to adjust the session token duration

- Search all files and replace three instances of '4h' with '1m' so that the session will expire after 1 minute
- In backend perform: "rm -r node_modules; npm install; npm run dev" to rebuild
- Login and wait for just over a minute, then attempt to navigate to another page or create a post
- You should be navigated back to the login page with a 401 error

Closes #651

### DIFF
--- a/backend/src/api/auth.ts
+++ b/backend/src/api/auth.ts
@@ -38,7 +38,7 @@ router.post('/login', async (req, res) => {
   }
 
   const user = userToToken(foundUser);
-  const token = sign(user, getJWTSecret(), { expiresIn: '2h' });
+  const token = sign(user, getJWTSecret(), { expiresIn: '4h' });
   const expiresAt = addHours(2);
 
   res.status(200).send({ token, user, expiresAt });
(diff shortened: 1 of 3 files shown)


markiianbabiak (changes requested):
Screenshot 2025-02-03 at 2 45 25 AM
This is the page I got instead of the login page.

To recreate:
Login, go to any board, go to canvas, wait, try switching to workspace.


Side note: consider also showing a pop-up message at the top saying "Session expired. Redirected to login page"

PR #661, opened 2025-02-11
572 add individual work option to generative task workflow

### LINKED ISSUE #572: Add "Individual Work" Option to Generative Task Workflow
### Description
Add an additional option for Individual Work to the Generation Task Workflow that involves selecting only a single group whereby all members of that group must each complete the assigned task.
Screenshot 2023-11-30 at 3 44 58 PM

**Details**
This will add an "Individual Work" option to the Generation Task workflow (under the Create Workflow tab). When selected, the "Assigned Groups" (multi-selection) will be replaced with an "Assigned Members" (single-selection). The current Generation Task workflow involves sending the same generation task to all members of a group (and is marked as complete when any member of the group completes the work), but with Individual Work each member receives their own copy of the task and is only completed when that individual completes the work. 

**Tasks**

- [x] Add a new radio button group with options for "Group Work (assigned to each group)" and "Individual Work (assigned to each member", under the "Destination" selector.
- [x] When Individual Work is selected, replace the "Assigned Groups" multi-group selector with an "Assigned Members" (single-group selector; as a required field)
- [x] When Group Work is selected, the "Assigned Groups" multi-group selector should be displayed and the Workflow and Workspace should behave as before
- [x] Running the Individual Work, Generation Task workflow should assign a copy of the task to each individual member of the selected single group
- [x] Let's remove the Pending Tasks section in the Workspace sidebar (as shown below) and have all new tasks added directly to Active Tasks (we can either continue to use the "Begin Task" button for starting a new task or use only the "Show/Hide Task" buttons for Active Tasks once the Pending Tasks and Active Tasks are merged)
Screenshot 2023-11-30 at 4 57 42 PM

- [x] Each individual is responsible for completing their own Individual Work tasks in the Workspace, this is denoted by replacing the "Group [#]: [name]" field in the Workspace sidebar with "Individual: [workflow name]"
Screenshot 2023-11-30 at 4 19 22 PM

- [ ] Group tasks are currently denoted in the sidebar with "Group [#]: [student prompt]", let's swap that with "Group [#]: [workflow name]

-  Let's clean up the UI with a bunch of small changes, as seen below 
Screenshot 2023-11-30 at 4 42 04 PM

- [x] Remove the group number from the workflow title (at the top of the page) and set the font-weight to bold and the margin to 15px (all directions)
- [x] For the prompt, switch it to a paragraph element, set the font-size to 18px, and set the margin to 15px (all directions)
- [x] Move the task (e.g., Create 1 Post) to underneath progress bars with the text "Task:" above it
- [ ] "Group Members" should be replaced with "Group [#]:"
- [x] The "Task:" and "Group [#]:" text should switched to paragraph elements
- [x] For Individual Work tasks, the "Group [#]:" label and members list can be removed (as seen below)
Screenshot 2023-11-30 at 4 55 15 PM

markiianbabiak:
## Details
Changes to ck-workspace and ck-monitor to add an individual task option
- 

Closes #572


JoelWiebe (commented):
I just noticed a few things, but we don't necessarily need to address them all:

- If a student is added to a group after that group's individual work task is assigned, they won't have access to the individual work task
- The progress bar still refers to the Avg Group Member's Progress, maybe we can change this to Avg Student Progress
- Posts added to the canvas are added directly on top of each other rather than slightly offset from one another (i.e., if an individual work task involves 3 posts, the first second and third post could be slightly vertically and horizontally offset from the previous pst)
- Posts added to the canvas are not displayed immediately; canvas must be reloaded to display new posts

Do any of these sound feasible to quickly fix?

markiianbabiak:
@JoelWiebe 
- I'll take a look at that
- But group members are not necessarily students, teachers can be part of groups too. So if the problem is naming I would leave it as is
- The last point is resolved in the workspace socket events, therefore, I will fix the previous one there too.

JoelWiebe:
Thanks. No rush for the run, but maybe we can rename that to Avg Member Progress for Individual Work -- but this can wait until later.

PR #674, opened 2025-03-06
Fix COIL event bugs

### LINKED ISSUE #672: BUG: Post-COIL Bugs
### Bug
**Overview**
Items that came up in preparation for the COIL event

**Proposed tasks to fix**
- [ ] User account creation enforces unique display names rather than unique emails
- [ ] Reset password link has invalid prefix on production server, but not on localhost
- [ ] Workspace doesn’t display bottom of view if displaying a long post post
- [ ] On mobile device, dashboard is zoomed in on login (on some devices)
- [ ] In buckets view, posts are often dragged when trying to scroll, therefore, disable dragging of posts

JoelWiebe:
**Completed**

- Prevent dragging of posts between buckets in CK Buckets component
- Allow workspace to scroll for large posts
- Added https to password reset link if http is not already specified in the server name
- Cleared email address field when forgot password reset link is requested
- Added https to password reset link if http is not already specified in the server name
- Cleared email address field when forgot password reset link is requested
- Removed enforcement of unique display name
- Set enforcement of unique email
- Display email already in use error if appropriate on registration

**Test**

- [x] Cannot create a new user if email already exists
- [x] Verify that email already exists error is appropriately displayed for above case
- [x] Can create a new user with the same display name as another user
- [x] On forgot password reset link request, verify that email field is cleared after request is made
- [x] Verify that link works on local host and will need to test again in the future on the production server
- [x] Verify that in CK Buckets component, posts can no longer be dragged between buckets but can still be opened
- [x] Verify on Workspace that when post contents are very long that the user can scroll down to still see the group list below.

Closes #672

### DIFF
--- a/backend/src/api/auth.ts
+++ b/backend/src/api/auth.ts
@@ -58,8 +58,10 @@ router.post('/login', async (req, res) => {
 router.post('/register', async (req, res) => {
   const body = req.body;
 
-  const exists = await dalUser.findByUsername(body.username);
-  if (exists) return res.sendStatus(400);
+  const exists = await dalUser.findByEmail(body.email);
+  if (exists) {
+    return res.status(400).send({message: 'Email already in use.'});
+  }
 
   const savedUser = await dalUser.create(body);
 
--- a/backend/src/api/auth.ts
+++ b/backend/src/api/auth.ts
@@ -106,7 +106,11 @@ router.post('/forgot-password', async (req, res) => {
     }
 
     // 5. Send an email to the user with a link containing the token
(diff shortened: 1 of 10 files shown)


markiianbabiak:
Some posts in the workspace look a little weird.
1. The borders are a little weird here
Screenshot 2025-03-06 at 11 49 07 AM
2. When a post is just one long string without spaces 
Screenshot 2025-03-06 at 11 49 19 AM

JoelWiebe:
That's super strange. There shouldn't be any multi colored posts anymore. Can you try rebuilding the front-end?

PR #691, opened 2025-06-10
Added display attributes for bucket posts

### LINKED ISSUE #690: BUG: Fix post creation with no display position
### Bug
**Overview**
It appears that posts created in the workspace for a bucket may not be getting locational coordinates needed by the post object and for displaying if moved to the canvas.

We need to ensure that default coordinate values are always given to posts created in the workspace.

JoelWiebe:
Added display attribute coordinates for posts created in CK Workspace with bucket destinations

## TODO

- Test that posts are created in ck workspace with bucket destination without error
- Confirm that if these posts are moved to the canvas that they have position coordinates on the canvas
- Verify that this resolved the previous issue with displayAttributes error in dalPost.ts

Closes #690


JoelWiebe:
@marieklinaeva Thanks! Yes, the default behaviour that existed for CK Workspace when creating posts directly to the canvas was to stack them, so I just applied this to posts created for the buckets as well.

Let's fix this behaviour in a future issue.

